PR #3, opened 2021-03-09
Topic based schedule updates

gbiggs:
### Implemented feature

The existing method of updating schedule mirrors via a pull-based service request is not scalable to large numbers of mirrors. This group of PRs changes schedule mirroring to use push-based communication via topics.

There are three related PRs that must be merged together:
1. This PR,  which adds the topic-based communication functionality.
2. [rmf_traffic#5](https://github.com/open-rmf/rmf_traffic/pull/5), which makes the mirrors more robust to duplicate changes arriving, at the cost of efficiency (this is a temporary work-around).
3. [rmf_internal_msg#3](https://github.com/open-rmf/rmf_internal_msgs/pull/3) which adds the necessary messages.

Additionally, [this PR adding equality comparison for `Query` objects](https://github.com/open-rmf/rmf_traffic/pull/3) is a dependency.

The schedule node previously provided a service to pull updates from the schedule database. Mirrors would call this service when they wanted the recent changes in the schedule. Usually this was done in response to a broadcast from the schedule node informing them that changes are available; the mirrors would generally sleep until woken by this broadcast.

The functionality added in these PRs changes communication of schedule updates from pull to push. The service and the broadcast changes-available notifier are removed. Instead, when a query is registered with the schedule node, it creates a topic named `query_updates_` with the query ID appended, and responds to the query registration call with the ID of the registered query. The mirror manager node knows to subscribe to the correct topic based on the received query ID, and does so. The schedule node then broadcasts any changes relevant to that query over that topic, allowing the mirror to receive the changes without needing to make a request first.

For further scalability improvements, when a query is registered that is the same as an existing query, the schedule node does not create a new topic. Instead it sends the ID of the existing query. This allows multiple mirrors to listen to the same topic if they are interested in the same query.

To allow a new mirror to receive the entire database even though it joins after some changes have been broadcast, a new mirror joining an existing query will cause the entire database to be re-broadcast to all mirrors of that query. This mostly ensures that all mirrors have all information. However, there are two caveats to this that require temporary workarounds.

1. Delays in topic subscriptions being connected means that sometimes the whole-database message is missed by the new query mirror. This causes it to fail to interpret subsequent individual-change messages, such as a delay message for a participant that it doesn't know about or the erasure of a route that it doesn't have yet. To work around this, encountering an error in interpreting a patch will cause that mirror to request the full database again.
2. Existing mirrors will receive participants, changes, etc. that they already know about. This was previously considered an error. To work around this, these are no longer treated as errors. A log message is output and the duplicate entry is ignored.

These two workarounds will no longer be necessary once how schedules are synchronised is updated to split participant information synchronisation and itinerary information synchronisation. This will be done in a future PR.

### How to test

Start the office demo:

```
ros2 launch rmf_demos office.launch.xml
```

In another terminal, trigger a couple of loop requests:

```
ros2 run rmf_demos_tasks dispatch_loop -s coe -f lounge -n 3 --use_sim_time
```

The schedule should be displayed as normal in `rviz`.

The schedule information flowing over the topic can be verified:

```
ros2 topic echo /rmf_traffic/query_update_1
```

Finally, killing the node that is visualising the schedule in rviz and restarting it should show the schedule disappear when the node is killed and then get displayed again after restarting it.

### DIFF
--- a/rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
+++ b/rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
@@ -33,8 +33,8 @@ const std::string UnregisterParticipantSrvName = Prefix +
   "unregister_participant";
 const std::string RegisterQueryServiceName = Prefix + "register_query";
 const std::string UnregisterQueryServiceName = Prefix + "unregister_query";
-const std::string MirrorUpdateServiceName = Prefix + "mirror_update";
-const std::string MirrorWakeupTopicName = Prefix + "mirror_wakeup";
+const std::string QueryUpdateTopicNameBase = Prefix + "query_update_";
+const std::string RequestChangesTopicName = Prefix + "request_changes";
 const std::string ScheduleInconsistencyTopicName = Prefix +
   "schedule_inconsistency";
 const std::string NegotiationAckTopicName = Prefix +
--- a/rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/MirrorManager.hpp
+++ b/rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/MirrorManager.hpp
@@ -76,13 +76,9 @@ public:
 
   /// Attempt to update this mirror immediately.
   ///
-  /// \param[in] wait
-  ///   How long to block the current thread while waiting for the mirror to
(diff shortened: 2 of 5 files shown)


mxgrey:
This can be closed in favor of #17

Issue #4, opened 2021-03-09
Pybinding functions that return raw pointers causes Segfaults


aaronchongth:
https://github.com/open-rmf/rmf_ros2/blob/f5e11306baf04c26cdbf2876e249e521e46d8cf8/rmf_fleet_adapter_python/src/graph/graph.cpp#L171-L174

https://github.com/open-rmf/rmf_traffic/blob/main/rmf_traffic/include/rmf_traffic/agv/Graph.hpp#L512-L515

This is due to `pybind11` copying and deleting the raw pointer when it goes out of scope.

PR #5, opened 2021-03-09
Adding reference_internal tag to function bindings that return raw pointers

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

### Fixed bug

https://github.com/open-rmf/rmf_ros2/issues/4

### Fix applied

The functions that were bound returns a raw pointer, using the fallback `return_value_policy::take_ownership`, it deletes the item after it goes out of scope in python. Specifying them to use `return_value_policy::reference_internal` prevents that.


Yadunund (approved):
Tested with hardware ✔️

aaronchongth:
Closing due to unsigned commits.

PR #11, opened 2021-03-18
Feature/python binding planner

Yadunund:
### Implemented feature
Added additional python bindings:
1. Minimal support for `rmf_traffic::agv::Planner`. A `get_plan_waypoints` function is implemented which returns a list of waypoints in the plan if the planner was successful.
2. Updated the Node wrapper to allow users to set `use_sim_time` parameter to `true`. This is helpful for testing fleet adapters in simulation.

### DIFF
--- a/rmf_fleet_adapter_python/src/planner/planner.cpp
+++ b/rmf_fleet_adapter_python/src/planner/planner.cpp
@@ -7,8 +7,6 @@
 #include "rmf_traffic_ros2/Time.hpp"
 #include <rmf_traffic/agv/Planner.hpp>
 
-#include <stdexcept>
-
 namespace py = pybind11;
 
 using Plan = rmf_traffic::agv::Plan;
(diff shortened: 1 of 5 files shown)


rmf_fleet_adapter_python/src/planner/planner.cpp
```diff
@@ -144,4 +153,44 @@ void bind_plan(py::module &m) {
...
+           py::arg("graph"),
+           py::arg("traits"))
+      .def_property_readonly("graph", &Configuration::graph)
+      .def_property_readonly("traits", &Configuration::vehicle_traits);
+  
+  // Options =============================================================
+  // TODO
+  // Light wrappers
+  py::class_<Interpolate::Options>(m, "InterpolateOptions")
+      .def(py::init<>());
```
tanyouliang95, line 169: Looks fine to me. One minor point, should we just remove binding for `Interpolate::Options` for now since it is not being used in the `Configuration` yet

tanyouliang95 (changes requested):

rmf_fleet_adapter_python/src/planner/planner.cpp
Yadunund: Done https://github.com/open-rmf/rmf_ros2/pull/11/commits/e9ecb827bd33f0668d80cbda7d02ca5c56f4f76f

tanyouliang95 (approved):

tanyouliang95 (approved):

Issue #12, opened 2021-03-20
full_control fleet adapter will crash if empty StartSet is passed into FleetUpdateHandle::add_robot()


Yadunund:
### Required information:

- Operating system and version: Ubuntu 20.04
  - 
- OpenRMF installation type: Source
  - 
- OpenRMF version or commit hash main
  - 
- ROS distribution and version: ROS 2 Foxy
  - 
- ROS installation type: Binaries
  - 
- Package or library, if applicable:
  - 

### Description of the bug

As summarized [here](https://github.com/osrf/rmf_core/issues/315)

PR #13, opened 2021-03-22
Handle empty startset in full_control

Yadunund:

### Fixed bug

Fix for https://github.com/open-rmf/rmf_ros2/issues/12

### Fix applied
Check if output of `compute_plan_starts() is empty and if so return after outputting an error

### DIFF
--- a/rmf_fleet_adapter/src/full_control/main.cpp
+++ b/rmf_fleet_adapter/src/full_control/main.cpp
@@ -474,12 +474,15 @@ struct Connections : public std::enable_shared_from_this<Connections>
     {
       RCLCPP_ERROR(
         adapter->node()->get_logger(),
-        "Unable to compute a StartSet for robot [%s]. This can happen if the "
-        "level_name in the RobotState message does not match any of the "
-        "map names in the navigation graph supplied or if the location "
+        "Unable to compute a StartSet for robot [%s] using level_name [%s] and "
+        "location [%f,%f,%f] specified in its RobotState message. This can "
+        "happen if the level_name in the RobotState message does not match any "
+        "of the map names in the navigation graph supplied or if the location "
         "reported in the RobotState message is far way from the navigation "
-        "graph. This robot will not be added to the fleet [%s]",
+        "graph. This robot will not be added to the fleet [%s].",
         state.name.c_str(),
+        state.location.level_name.c_str(),
+        l.x, l.y, l.yaw,
         fleet_name.c_str());
 
         return;
--- a/rmf_fleet_adapter/src/full_control/main.cpp
+++ b/rmf_fleet_adapter/src/full_control/main.cpp
@@ -476,7 +476,7 @@ struct Connections : public std::enable_shared_from_this<Connections>
       RCLCPP_ERROR(
         adapter->node()->get_logger(),
         "Unable to compute a StartSet for robot [%s] using level_name [%s] and "
-        "location [%f,%f,%f] specified in its RobotState message. This can "
+        "location [%f, %f, %f] specified in its RobotState message. This can "
         "happen if the level_name in the RobotState message does not match any "
         "of the map names in the navigation graph supplied or if the location "
         "reported in the RobotState message is far way from the navigation "
--- a/rmf_fleet_adapter/src/full_control/main.cpp
+++ b/rmf_fleet_adapter/src/full_control/main.cpp
@@ -467,11 +467,26 @@ struct Connections : public std::enable_shared_from_this<Connections>
           path_request_pub, mode_request_pub);
 
     const auto& l = state.location;
+    const auto& starts = rmf_traffic::agv::compute_plan_starts(
+        *graph, state.location.level_name, {l.x, l.y, l.yaw},
+        rmf_traffic_ros2::convert(adapter->node()->now()));
+    if (starts.empty())
+    {
+      RCLCPP_ERROR(
+        adapter->node()->get_logger(),
+        "Unable to compute a StartSet for robot [%s]. This can happen if the "
+        "level_name in the RobotState message does not match any of the "
+        "map names in the navigation graph supplied or if the location "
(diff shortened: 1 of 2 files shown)


rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -467,11 +467,26 @@ struct Connections : public std::enable_shared_from_this<Connections>
           path_request_pub, mode_request_pub);
 
     const auto& l = state.location;
+    const auto& starts = rmf_traffic::agv::compute_plan_starts(
+        *graph, state.location.level_name, {l.x, l.y, l.yaw},
+        rmf_traffic_ros2::convert(adapter->node()->now()));
+    if (starts.empty())
+    {
+      RCLCPP_ERROR(
```
mxgrey, line 475: I believe this will end up getting printed every time a new state update comes in, until the problem is resolved. It might be good to have something keep track of which robots this message has been printed out for, and only do it once.

rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -467,11 +467,26 @@ struct Connections : public std::enable_shared_from_this<Connections>
...
+        adapter->node()->get_logger(),
+        "Unable to compute a StartSet for robot [%s]. This can happen if the "
```
mxgrey, line 477: Let's also print out what we received for the robot's coordinates and map name. That may help users with debugging.

mxgrey (changes requested):

rmf_fleet_adapter/src/full_control/main.cpp
Yadunund: Actually, it will only be printed once as the `Connections::add_robot()` function is only called the first time the `RobotState` message for that robot is received. I've tested this by having the `slotcar` continuously publish `RobotState` msgs with invalid `level_name`. 
https://github.com/open-rmf/rmf_ros2/blob/87de246ade6c371b6d594080a34b971721c55085/rmf_fleet_adapter/src/full_control/main.cpp#L746-L752

rmf_fleet_adapter/src/full_control/main.cpp
Yadunund: Good point. Done https://github.com/open-rmf/rmf_ros2/pull/13/commits/0818cd5f2046d08675778391762f7931aa0e2db8

rmf_fleet_adapter/src/full_control/main.cpp
mxgrey: Ah, I see.

In that case, it means if the initial state for the robot was invalid, then the robot can never be added until the fleet adapter is restarted, which also does not seem great to me. But then again, this is a legacy adapter for an API that we don't really want people to be using anyway, so I'm okay with leaving it less robust than it could be.

rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -467,11 +467,29 @@ struct Connections : public std::enable_shared_from_this<Connections>
...
     const auto& l = state.location;
+    const auto& starts = rmf_traffic::agv::compute_plan_starts(
+        *graph, state.location.level_name, {l.x, l.y, l.yaw},
+        rmf_traffic_ros2::convert(adapter->node()->now()));
+    if (starts.empty())
+    {
+      RCLCPP_ERROR(
+        adapter->node()->get_logger(),
+        "Unable to compute a StartSet for robot [%s] using level_name [%s] and "
+        "location [%f,%f,%f] specified in its RobotState message. This can "
```
mxgrey, line 478: Nitpick:

```suggestion
        "location [%f, %f, %f] specified in its RobotState message. This can "
```

Floating point numbers can be hard to visually recognize without any spacing between them.

mxgrey:
One other suggestion for this PR: Let's have the `add_robot` function throw an exception if it receives an empty start set. This isn't just an issue for the legacy fleet adapter. Any fleet adapter that gives us an empty start set is going to have problems.

rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -467,11 +468,31 @@ struct Connections : public std::enable_shared_from_this<Connections>
...
+        "happen if the level_name in the RobotState message does not match any "
+        "of the map names in the navigation graph supplied or if the location "
+        "reported in the RobotState message is far way from the navigation "
+        "graph. This robot will not be added to the fleet [%s].",
+        state.name.c_str(),
+        state.location.level_name.c_str(),
+        l.x, l.y, l.yaw,
+        fleet_name.c_str());
+
+        throw std::runtime_error(
```
mxgrey, line 489: This isn't what I meant when I suggested throwing an exception. I think it's fine to just let it continue from here and ignore the robot in the future.

Effectively what I meant is that [this assertion](https://github.com/open-rmf/rmf_ros2/blob/888a35a418a623e9e063857b7d54d3cd18fdbd50/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L987) should be promoted to throwing an exception, so the robot adding routine always halts there no matter what mode it was built in.

Yadunund: Aaah I apologise for misunderstanding your comment. I hope the latest change reflects what you had in mind.

Achllle:
I can confirm this fixes [osrf/rmf_core/#266](https://github.com/osrf/rmf_core/issues/266). It would be useful to inform the user why exactly the start set is empty.

mxgrey:
@Achllle I've [added some more information](https://github.com/open-rmf/rmf_ros2/pull/13/commits/ea4de7cb04971def7f763833bfe71d275bc5beed) to the error message that may help debug the problem.

mxgrey (approved):

Achllle:
Amazing, those debug messages are very helpful @mxgrey

PR #15, opened 2021-03-24
Lane closure

mxgrey:
### Implemented feature

Allow fleet adapters to close navigation graph lanes.

The `rmf_fleet_adapter::agv::FleetUpdateHandle` API can now be told to close or open navigation graph lanes during runtime.

For the legacy fleet adapter (aka the `full_control` adapter) you can send a `LaneRequest` message to open or close lanes at any time. If the `full_control` adapter detects that the robot is currently on a lane that was just closed, then it will have the robot retreat back to the start of the lane and find a new plan from there. **NOTE**: We do not offer very strong guarantees about the effectiveness or reliability of this behavior for the legacy fleet adapter, because the `rmf_fleet_msgs` API does not always convey as much information as we need to make the behavior reliable.

### Dependencies

- [x] https://github.com/open-rmf/rmf_internal_msgs/pull/5
- [x] https://github.com/open-rmf/rmf_traffic/pull/11

### DIFF
--- a/rmf_fleet_adapter/CHANGELOG.rst
+++ b/rmf_fleet_adapter/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog for package rmf_fleet_adapter
 
 1.3.0 (2021-XX-XX)
 ------------------
+* Add API for opening and closing lanes: [#15](https://github.com/open-rmf/rmf_ros2/pull/15)
+    * Added `open_lanes` and `close_lanes` CLI tools for issuing requests
 * Allow Traffic Light APIs to update the location of a robot while it is idle: [#270](https://github.com/osrf/rmf_core/pull/270)
 * Allow TrafficLight and EasyTrafficLight API to update battery level: [#263](https://github.com/osrf/rmf_core/pull/263)
 * Migrating to a task dispatcher framework: [#217](https://github.com/osrf/rmf_core/pull/217)
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -319,6 +319,38 @@ target_include_directories(task_aggregator
 
 # -----------------------------------------------------------------------------
 
+add_executable(open_lanes src/open_lanes/main.cpp)
+
+target_link_libraries(open_lanes
+  PRIVATE
+    ${rmf_fleet_msgs_LIBRARIES}
+    rclcpp::rclcpp
+    rmf_fleet_adapter
+)
+
+target_include_directories(open_lanes
+  PRIVATE
+    ${rmf_fleet_msgs_INCLUDE_DIRS}
+)
+
+# -----------------------------------------------------------------------------
+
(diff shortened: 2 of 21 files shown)


mxgrey:
I've added some CLI tools for opening and closing lanes.

Using https://github.com/open-rmf/rmf_demos/pull/13 it's easy to test the lane closure feature with the following commands (using a separate terminal for each):

```
$ ros2 launch rmf_demos office.launch.xml
$ ros2 run rmf_demos_tasks dispatch_loop -s tinyRobot1_charger -f pantry -n 20 --use_sim_time
```

Once that is running, you can use

```
$ ros2 run rmf_fleet_adapter close_lanes tinyRobot <lanes>
```

to close `<lanes>` (replace with the indices of the lanes you want to close) and then use

```
$ ros2 run rmf_fleet_adapter open_lanes tinyRobot <lanes>
```

to reopen `<lanes>`.

For this demo, you can use the illustration below to pick which lanes to open and close and watch the result. Note that each "lane" is unidirectional, so closing lane 30 will block the robot from going east while closing lane 31 will block the robot from going west.
office_lanes_annotate

If you close the lane that the robot is currently on, the fleet adapter should instruct it to return to the start of the lane and find a new path from there.

Yadunund (approved):
Tested and works well! I know this is expected behaviour but when a robot is moving towards a `goal` and all lanes to that `goal` are closed, the `Planner` will fail to find a solution and `full_control` will crash:
```[full_control-15] [SearchForPath] CRITICAL ERROR: Impossible plan requested! Participant [tinyRobot_1] owned by [tinyRobot] Requested path (4) --> (2)```

Just wondering if the fleet adapter should do some kind of check to see if the requested lane closure will lead to this type of error and hence not close the lane?

I was also planning to open a ticket in `rmf_demos` to add scripts to open/close lanes but I see that they have been added here directly 👍🏼

mxgrey:
> Just wondering if the fleet adapter should do some kind of check to see if the requested lane closure will lead to this type of error and hence not close the lane?

I think the fleet adapter **needs** to respect lane closing commands, because it may be the case that sending the robot down a closed lane would be harmful to the robot or others.

Instead we need to improve the error handling pipeline for critical task failures. I've added a ticket #16 for this.

Yadunund:
Agreed, that is a better approach.

Also, maybe in a future PR we can have `FleetUpdateHandle` publish to `/closed_lanes` (akin to how it publishes to `/fleet_states`)? Assuming we have fleets integrated with the new fleet adapter APIs, this information will not be broadcasted.

mxgrey:
> Also, maybe in a future PR we can have FleetUpdateHandle publish to `/closed_lanes` (akin to how it publishes to `/fleet_states`)? Assuming we have fleets integrated with the new fleet adapter APIs, this information will not be broadcasted.

Yeah, I wanted to save that for later to give us time to decide for sure how we want to approach that before committing to that particular topic/message long-term.

Yadunund (approved):

Issue #20, opened 2021-04-01
restart error Failed to correctly load participant registry


txlei:
Hi, I am not sure why this error happens after restarting the office demo. I've done a PC restart, rosdep update and rebuild. Currently doing a clean rebuild. 
```
[rmf_traffic_schedule-1] [FATAL] [1617249496.967795104] [rmf_traffic_schedule_node]: Failed to correctly load participant registry: yaml-cpp: error at line 5208, column 5: Participant description missing [profile] field
[rmf_traffic_schedule-1] 
[rmf_traffic_schedule-1] terminate called after throwing an instance of 'std::runtime_error'
[rmf_traffic_schedule-1]   what():  yaml-cpp: error at line 5208, column 5: Participant description missing [profile] field
[ERROR] [rmf_traffic_schedule-1]: process has died [pid 42702, exit code -6, cmd '/home/leia/rmf_demos_ws/install/rmf_traffic_ros2/lib/rmf_traffic_ros2/rmf_traffic_schedule --ros-args --params-file /tmp/launch_params_ttpvadv4'].
[rviz2-7] [INFO] [1617249497.414677664] [door_panel_requester_node]: New door [hardware_door] found, refreshing...
[rviz2-7] [INFO] [1617249497.416391942] [door_panel_requester_node]: New door [main_door] found, refreshing...
[rviz2-7] [INFO] [1617249497.632252148] [door_panel_requester_node]: New door [coe_door] found, refreshing...
[...]

PR #22, opened 2021-04-06
Fix a bug that made the participant registry grow

mxgrey:
### Fixed bug

The code path for loading participants from a file intersects with the code path for registering new participants, and in both cases the new participant was being written to file. This is a problem because now the log file will grow simply by loading it.

### Fix applied

Split the intersecting code path based on whether we are currently reading from the log or not.


arjo129 (approved):
Looks good to me. Seems to work as described.

mxgrey:
@txlei I can't be sure, but this could be related to #20

PR #23, opened 2021-04-07
Test/enforce logsize remains same

arjo129:
# Adds a test to prevent the regression fixed in #22

Adds a one line test to ensure that log file size remains the same across runs. This is to prevent regressions like the one fixed in #22.


mxgrey (approved):
I've confirmed that this test fails prior to #22 and passes now. Perfect regression test :+1:

PR #24, opened 2021-04-08
refactor task components in fleet adapter

youliangtan:
This is a broken down pr from this feedback https://github.com/osrf/rmf_core/pull/302#issuecomment-786484813

- Previously the task summary msg is not fully populated, as task information (`taskprofile`) is missing. this is addressed here.
- Refactor work in `taskManager` and `fleetupdatehandle`

Note: let me know if the refactoring work makes sense in making the implementation clearer. The expected behavior is that this won't break the existing demo scenarios.

Signed-off-by: youliang <tan_you_liang@hotmail.com>


Yadunund:
@tanyouliang95 could you explain the motivation behind the refactoring of `TaskManager` and `FleetUpdateHandle` along with a summary of the changes?

tanyouliang95:
This PR is mainly on refactoring work, along with changing some public fn in `TaskManager`, and removing some of the redundant vars. Also, some change is made to make it supportable for future task revival (without ignoring submitted bidding task).

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -633,28 +561,20 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
       return;
     }
 
-    auto request_to_cancel_it = assigned_requests.find(id);
-    if (request_to_cancel_it == assigned_requests.end())
+    if (!is_queued(id))
```
Yadunund, line 564: The proposed change alters the intention of the code. The original approach checks for whether or not the task to be cancelled was assigned to this fleet (whether queued or active). However the new approach only checks for queued tasks. If the task is currently active, it will exit after printing ```Unable to cancel task with task_id:[%s] as it is not queued to fleet:[]```. But this is not the case. We would want the execution to reach the block ```RCLCPP_WARN(
        node->get_logger(),
        "Unable to cancel active task with task_id:[%s]. Only queued tasks may "
        "be cancelled.",
        id.c_str());``` below

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -718,23 +635,50 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
 auto FleetUpdateHandle::Implementation::is_valid_assignments(
   Assignments& assignments) const -> bool
 {
-  std::unordered_set<std::string> executed_tasks;
```
Yadunund, line 721: The original intention of this function is to ensure that none of the tasks in the previously generated assignments has begun execution. The was achieved efficiently in the original implementation. With your changes, you call `is_queued()` inside a nested `for` loop which in turn runs two `for` loops to compile a list of `queued_tasks`. Repeating this for assignment per agent is highly inefficient. Please retain the original implementation.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -718,23 +635,50 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
...
+      assignment_size++;
     }
   }
 
-  return true;
+  // Also check if size of total queued tasks is the same as assignments
+  for (const auto& mgr : task_managers)
+    queued_task_size += mgr->task_queue().size();
+
+  return (queued_task_size == assignment_size);
```
Yadunund, line 655: I do not think there is any need for this comparison.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -718,23 +635,50 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
...
+}
+
+//==============================================================================
+bool FleetUpdateHandle::Implementation::is_queued(
```
Yadunund, line 659:
```suggestion
bool FleetUpdateHandle::Implementation::is_task_queued(
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -718,23 +635,50 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
...
+  const std::string& task_id ) const
+{
+  std::unordered_set<std::string> queued_tasks;
```
Yadunund, line 662: Constructing a complete set of `queued_tasks` and then returning ```return (queued_tasks.find(task_id) != queued_tasks.end());``` is a bit inefficient. This way in the best case the function returns a lot earlier.

Instead you can have 

```
for (const auto& [context, mgr] : task_managers)
{
  for (const auto& request : mgr->requests())
  {
    if (request->id() == task_id)
      return True
  }
}
return false
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -718,23 +635,50 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
...
+  {
+    for (const auto task : mgr->task_queue())
+      queued_tasks.insert(task->id());
+  }
+
+  return (queued_tasks.find(task_id) != queued_tasks.end());
+}
+
+//==============================================================================
+bool FleetUpdateHandle::Implementation::is_current(
```
Yadunund, line 673:
```suggestion
bool FleetUpdateHandle::Implementation::is_active_task(
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -718,23 +635,50 @@ void FleetUpdateHandle::Implementation::dispatch_request_cb(
...
+  const std::string& task_id ) const
+{
+  std::unordered_set<std::string> current_tasks;
```
Yadunund, line 676: A bit inefficient. Refactor as suggested for `is_queued_task()`

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -860,7 +804,7 @@ void FleetUpdateHandle::Implementation::publish_fleet_state() const
 //==============================================================================
 auto FleetUpdateHandle::Implementation::allocate_tasks(
   rmf_task::ConstRequestPtr new_request,
-  rmf_task::ConstRequestPtr ignore_request) const -> std::optional<Assignments>
+  std::string ignore_request_id) const -> std::optional<Assignments>
```
Yadunund, line 807: Revert to original signature.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -877,42 +821,32 @@ auto FleetUpdateHandle::Implementation::allocate_tasks(
...
   for (const auto& t : task_managers)
   {
-    states.push_back(t.second->expected_finish_state());
-    constraints_set.push_back(t.first->task_planning_constraints());
-    const auto requests = t.second->requests();
-    pending_requests.insert(
-      pending_requests.end(), requests.begin(), requests.end());
-  }
-
-  // Remove the request to be ignored if present
```
Yadunund, line 887: I'd suggest keeping this original implementation that does not require the internal task queue of the task manager to be exposed. Another benefit is that it calls `TaskManager::requests()` which automatically returns requests that are not charging one. In the future, as we implement other kinds of auto included tasks such as parking, we will only need to update the implementation of `TaskManager::requests()` without modifying this block.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -176,13 +172,14 @@ class FleetUpdateHandle::Implementation
...
-  std::unordered_map<std::string, Assignments> bid_notice_assignments = {};
+  
+  // Map to store task id with the profile msg of the received tasks
+  using TaskProfile = rmf_task_msgs::msg::TaskProfile;
+  std::unordered_map<std::string, TaskProfile> received_task_profiles = {};
+  
+  // Tuple which stored the latest received BidNotice
+  std::tuple<rmf_task::ConstRequestPtr, Assignments> latest_bid_notice_assignments;
 
-  std::unordered_map<
```
Yadunund, line 182: Hmm I think keeping track of these will be essential but we will need some ways to maintain a fixed size of these data structures.

rmf_fleet_adapter/test/tasks/test_Loop.cpp
```diff
@@ -178,17 +176,17 @@ SCENARIO("Test loop requests")
 
           ++completed_1_count;
         }
-        else
-          CHECK(false);
+        // else
```
Yadunund, line 179: Why are we commenting out tests instead of fixing them?

Yadunund (changes requested):
Here's my feedback. I found the refactor in some places to add more complexity or not achieve the intention of the original code. Generally also it would be good not to mix too many unrelated changes in the same PR. The refactor could have been in a separate PR and similarly for the changes introduced in this commit https://github.com/open-rmf/rmf_ros2/pull/24/commits/edab2601c26e9f88b549fe9eeb35c8f82b6140da which seems to 1) add helpers for a future feature (in which case these changes could have been introduced in the same PR as the future feature) and 2) "Fix" tests by commenting them out?

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -152,16 +152,12 @@ class FleetUpdateHandle::Implementation
   rmf_utils::optional<rmf_traffic::Duration> default_maximum_delay =
       std::chrono::nanoseconds(std::chrono::seconds(10));
 
-  AcceptDeliveryRequest accept_delivery = nullptr;
-  std::unordered_map<RobotContextPtr, std::shared_ptr<TaskManager>> task_managers = {};
-
+  std::vector<std::shared_ptr<TaskManager>> task_managers;
```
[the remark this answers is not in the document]
tanyouliang95: The `RobotContext` is already specified within the `TaskManager`, thus removing this redundant one.

rmf_fleet_adapter/test/tasks/test_Loop.cpp
tanyouliang95: Directly remove the comments. The removal is actually a fix, as the new `ResponsiveWait`'s TaskSummmary msg will result in a false check, which in this case is untrue.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
tanyouliang95: I have removed all local data structures (use for task caching) in `FleetUpdateHandle` that will keep appending indefinitely. The logic here is to retrieve those caches that are stored within each task_manager, reduces complexity.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -152,16 +152,12 @@ class FleetUpdateHandle::Implementation
   rmf_utils::optional<rmf_traffic::Duration> default_maximum_delay =
       std::chrono::nanoseconds(std::chrono::seconds(10));
 
-  AcceptDeliveryRequest accept_delivery = nullptr;
```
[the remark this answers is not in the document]
tanyouliang95: Since it is no longer in used, mind as well just remove it? Surely I can add it back for the next removal.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
tanyouliang95: Yes, this still preserves the intention in a clearer way. The potential problem by using `get_executed_tasks` is if an active task is being moved back to the queue (future implementation), it will still be considered in the `get_executed_tasks`.

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.hpp
```diff
@@ -60,8 +57,16 @@ class TaskManager : public std::enable_shared_from_this<TaskManager>
...
-  void set_queue(const std::vector<Assignment>& assignments);
+  /// task planner, arg 'task_profiles' is merely used for status publishing.
+  void set_queue(
+    const std::vector<Assignment>& assignments,
+    const std::unordered_map<std::string, TaskProfile>& task_profiles = {});
+
+  /// set a vector of tasks
+  void set_queue(const std::vector<std::shared_ptr<Task>>& tasks);
+
+  // get tasks in the queue
```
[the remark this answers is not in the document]
tanyouliang95: I believe this helper function `task_queue()` is quite helpful to provide more info to the user about all current queued tasks. . The idea here is to let class `Task` be an observation component of `TaskManager`.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -291,12 +290,18 @@ class FleetUpdateHandle::Implementation
   /// new request and while optionally ignoring a specific request.
   std::optional<Assignments> allocate_tasks(
     rmf_task::ConstRequestPtr new_request = nullptr,
-    rmf_task::ConstRequestPtr ignore_request = nullptr) const;
+    std::string ignore_request_id = "") const;
```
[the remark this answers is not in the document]
tanyouliang95: Essentially, the internal function is just checking for the task_id, thus the arg `ignore_request_id`. Using a `Request` will make it less efficient.

Yadunund (changes requested):
Sorry for taking so long with this review.

Despite the title, I think the main intention of this PR is to fully populate the `TaskSummary` messages published by the `TaskManager`. And the changes to Task.cpp/hpp & TaskManager.cpp/hpp accomplish this well (left a couple of additional comments).

But the changes in `internal_FleetUpdateHandle.hpp` and `FleepUpdateHandle.cpp` are a bit problematic and hence dragging this PR down. My main issue is with getting rid of cache variables `bid_notice_assignments`, `generated_requests`, `assigned_requests` and `cancelled_task_ids` and introducing `latest_bid_notice_assignments`. The changes effectively restrict the fleet adapter to process BidNotice & DispatchRequest messages in a sequential order. ie, after a BidNotice is received, we assume we will only support receiving the DispatchRequest for the same task. I think we should be going in the opposite direction, ie, BidNotice and Dispatch requests can be received in any order and if there are any conflicts, we replan assignments accordingly.

I understand the intention to get rid of cache variables as right now they can increase indefinitely. I think a more ideal solution would be to define a new data structure
```
struct TaskEntry
{
  TaskProfile profile;
  ConstRequestPtr request;
  Assignments assignments;
  bool assigned;
  bool cancelled;
};
```
and then store an `std::unordered_map<std::string, TaskEntry>` as a cache as opposed to all the different unordered maps we currently have. We can then periodically purge this unordered_map off entries that are no longer relevant. Something like what was done [here](https://github.com/osrf/rmf_core/blob/0f155e9efea5e82d0da6723c45ee54964eb3cef5/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Adapter.cpp#L126-L140).
I think the impending redesign of the `TaskManager` could address some of these concerns too.

The use of `get_task_profiles()` function instead of having an `unordered_map<std::string, TaskProfile>` adds more complexity as well. Again, I guess the intention was to avoid having variables that expand indefinitely, but looping through all the tasks in all the TaskManagers each time to compile a list of TaskProfiles seems inefficient compared to a map insertion and lookup.

The other changes to these files do not seem to offer any performance improvements/fix bugs and are largely "another valid way of achieving the existing result". I think it would be great to further discuss how to make the internal implementation clearer for all users/maintainers and refactor accordingly.

But in the interest of getting the TaskSummary messages correctly populated for `rmf-web`, could we rebase this PR with only the changes to Task/TaskManager files? Alternatively, we can open another PR for the correction and leave this here to discuss and improve the internal implementation. Let me know what you think.

youliangtan:
Will close this. According to review, features in this PR will be further cherry-picked to separate chucks of pr.

PR #27, opened 2021-04-11
Check if an active phase exists before trying to cancel it

mxgrey:
### Fixed bug

Due to race conditions that can arise from the order in which the main rxcpp worker processes work, it's possible that `Task::cancel()` can be called on a `Task` instance that has no active phase. That will result in a segfault.

This PR fixes that issue by simply checking if the active phase exists before attempting to cancel it.


tanyouliang95 (approved):
Looks good!

PR #30, opened 2021-04-13
Give a floating tail prediction to the Responsive Wait phase

mxgrey:
### Fixed bug

Robots may sometimes have difficulty navigating around or negotiating with Responsive Waiting robots because the Responsive Waiting phase always reports to the schedule that its robot is going to vanish in a finite amount of time.

### Fix applied

In this PR, we add a "floating tail" to the schedule of the idle Responsive Wait robots. This floating tail will keep moving itself forward in time so that any robots that were planning on waiting for the Responsive Waiter to vanish will quickly end up in a conflict, and they will be forced to negotiate.


cnboonhan:
LGTM! It's not "perfect"  as a ResponsiveWait robot is less likely to move ( it seems ), but its very much preferable to having two robots stuck in the same place.

cnboonhan (approved):
LGTM! It's not "perfect" as a ResponsiveWait robot is less likely to move ( it seems ), but its very much preferable to having two robots stuck in the same place.

PR #32, opened 2021-04-14
Fix symlink-install compilation

tanyouliang95:
### Issue

This fix is partly to address the potential issue when build with `--symlink-install`. https://github.com/open-rmf/rmf/issues/24. 

### Fix
Previously the `rmf_fleet_adapter_python` will invoke a cmake build from `setup.py`. This PR will remove the need of `setup.py`, and install the pkg via `ament_python_install_package()` in the CMakeList.

Also, the package `ament_python` is removed, and replaced with [pybind11_vendor](https://github.com/ros2/pybind11_vendor). This will be installed during `rosdep`.

### Test
```bash
# Compile, with and without `--symlink-install`
ros2 run rmf_fleet_adapter_python test_adapter.py
```

-- in-progress --
- Pending `colcon test`


luca-della-vedova:
This fixes the symlink issue for me as well, I guess we should get colcon test to work first? Does the rclpy [matching code](https://github.com/ros2/rclpy/blob/master/rclpy/CMakeLists.txt#L213-L302) help?

tanyouliang95:
Added the test https://github.com/open-rmf/rmf_ros2/pull/32/commits/86b4a7f7fe09e31042851586a2eeced3f50a4472
```bash
colcon test --packages-select rmf_fleet_adapter_python --event-handlers console_direct+
```
Last but not least, once this is merged, users will need to re-run `rosdep update & install` sequence to ensure `pybind11-vendor` is locally installed.

luca-della-vedova:
> Added the test [86b4a7f](https://github.com/open-rmf/rmf_ros2/commit/86b4a7f7fe09e31042851586a2eeced3f50a4472)
> 
> ```shell
> colcon test --packages-select rmf_fleet_adapter_python --event-handlers console_direct+
> ```
> 
> Last but not least, once this is merged, users will need to re-run `rosdep update & install` sequence to ensure `pybind11-vendor` is locally installed.

Awesome! I tried building / testing / sourcing and importing the python bindings and it's working.
I agree that needing `rosdep update` / `install` is not ideal but that shouldn't be an issue for binary builds, while for source builds we can point it out to users.

luca-della-vedova (approved):

PR #35, opened 2021-04-22
Initialise documentation

gbiggs:
### Implemented feature

This PR sets up documentation using Sphinx for prose and Doxygen for API documentation. The API documentation is integrated into the Sphinx-based documentation.

Sphinx generates documentation from reStructuredText files. The top-level file includes a link to API documentation. The API documentation is generated using Doxygen and then integrated using Breathe and Exhale. All of this is automated using a Makefile and the Sphinx configuration script.


docs/conf.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'rmf_ros2'
+copyright = '2021, Open Source Robotics Corporation'
+author = 'Open Source Robotics Corporation'
+
+# The full version, including alpha/beta/rc tags
+release = '1.0.0'
```
mxgrey, line 25: Should we consider making this a `.in` template file so that we can have cmake automatically generate it from the version number that we declare in cmake? Otherwise we'll need to remember to update this field with each version change.

mxgrey (approved):
I only have one remark, left as an inline comment.

Is there anything I can/should do to test these changes? Trying to run `make` in the `docs` directory didn't really get me anywhere.

mxgrey:
Does [this doc page](https://www.sphinx-doc.org/en/master/usage/markdown.html) indicate that it's possible to use sphinx with simplistic markdown instead of rST? If so, would it be possible for us to leverage that instead? Or is there something about rST that's necessary for other reasons?

PR #45, opened 2021-04-28
adding failover mode, not enabled by default

marcoag:
### Implemented feature

This feature adds failover mode to the `robot_state_aggregator` node. This mode is disabled by default. This mode needs the [stubborn_buddies](https://github.com/open-rmf/stubborn_buddies) library at compilation time and it has to be enabled at runtime through the `failover_mode` argument.

This feature uses the [stubborn_buddies](https://github.com/open-rmf/stubborn_buddies) libraries for the hearbeat and watchdog nodes. The linktime_composition.cpp is also heavily based on the one from `stubborn_buddies. 

This PR does not add a dependency to `stubborn_buddies`, failover mode will only be added if the library is found at compilation time, otherwise `robot_state_aggregator` will be compiled without failover mode. If `robot_state_aggregator` was compiled without failover mode and the user tries to select it at runtime the process will throw an error and exit.


rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
```diff
@@ -116,13 +175,4 @@ class RobotStateAggregator : public rclcpp::Node
 
 };
 
-int main(int argc, char* argv[])
```
arjo129, line 119: Knit: Since there is no `main()` in this file, I recommend renaming it to `RobotStateAggregator.cpp` to prevent any confusion and help with overall readability.

arjo129 (changes requested):
This is pretty awesome. I've tested this and it works pretty well once you get it to build. The main issue with the current tricks used is that `colcon` may or may not decide to build `stubborn_buddies` before the `RobotStateAggregator`. I recommend using an environment variable like `RMF_ENABLE_FAILOVER` to enable it as a dependency. 

You can add the following to your `package.xml`
`<depend condition="$RMF_ENABLE_FAILOVER == 1">stubborn_buddies</depend>`

I've also got a small knit about the `main.cpp` remaining `main.cpp` as there is no `main` function in it anymore. It's probably better to rename it.

rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
marcoag: I was trying to initially maintain minimal changes to this file but I agree much better this way.

marcoag:
I started adding the environment variable and then I realized that instead of the parameter during runtime we could also rely on this variable. That way we avoid having to turn on one switch at compile time (env variable) and another one at runtime (parameter), which seems confusing, what do you say @arjo129 ?

arjo129:
That sounds awesome! The only use case where it maybe a problem is if the user has built from source without `stubborn_buddies` and then expects fail over to work.

marcoag:
In that case the execution fails and shouts to either remove the ail over option or to build with the dependency. I guess there's not much else that can be done.

marcoag:
I ended up keeping the parameter because I thought of the use case in which someone might want to enable failover mode only for certain nodes. The env variable will help with the selection at compilation time.

arjo129 (approved):
It looks like its working now. I recommend checking once on your own machine before merging.

marcoag:
First commit is not gpg signed, creating a new PR to solve the problem: https://github.com/open-rmf/rmf_ros2/pull/54

PR #47, opened 2021-04-30
Fix lookup of nearest charging waypoint in full control fleet adapters

Yadunund:
### Fixed bug

The full control fleet adapter presently suffers from a nasty bug due to a logical error in determining the charging waypoint of a robot when it is added to the fleet. Inside the implementation of `FleetUpdateHandle::add_robot()`, the [get_nearest_charger()](https://github.com/open-rmf/rmf_ros2/blob/5cbef9bff4778f83d54fb4318008f7e169706540/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L1012-L1013) method is invoked to return the nearest charging waypoint given the Start of the robot and a list of charging waypoints in the navigation graph. 

The error lies in the list of charging waypoints. This list was incorrectly populated with all the waypoints in the graph as opposed to only those with the `is_charger()` property set `true`. As a result, the charging waypoint for a robot was always set to the waypoint nearest to the robot at the time. Hence, when the `ChargeBattery` task begins, the robot would head to this waypoint and wait indefinitely for its battery to charge up. The problem is silent as no messages are printed during the `WaitForCharge` phase of the task. This problem went undetected for so long given that in all our simulations, the robots spawn at their charging waypoints so by coincidence the nearest waypoint was always a charging waypoint. In other non-legacy fleet adapters, the [RobotUpdateHandle::set_charger_waypoint()](https://github.com/open-rmf/rmf_ros2/blob/5cbef9bff4778f83d54fb4318008f7e169706540/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp#L94) method was being called to correctly set the charging waypoint and therefore overwriting the value set in `FleetUpdateHandle::add_robot()`. 

### Fix applied

* The list of charging waypoints is now correctly populated. 
* An error is thrown if none of the waypoints in the graph is a charger location
* A descriptive message with the expected and current charging rates is printed periodically when the `WaitForCharge` phase is active.

This branch has was successfully tested with hardware.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -749,7 +749,7 @@ std::optional<std::size_t> FleetUpdateHandle::Implementation::get_nearest_charge
   for (const auto& wp : charging_waypoints)
   {
     const rmf_traffic::agv::Planner::Goal goal{wp};
-    const auto& planner_result = (*planner)->plan(start, goal);
+    const auto& planner_result = (*planner)->setup(start, goal);
     const auto ideal_cost = planner_result.ideal_cost();
     if (ideal_cost.has_value() && ideal_cost.value() < min_cost)
     {
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -738,11 +738,13 @@ auto FleetUpdateHandle::Implementation::is_valid_assignments(
 }
 
 //==============================================================================
-std::size_t FleetUpdateHandle::Implementation::get_nearest_charger(
+std::optional<std::size_t> FleetUpdateHandle::Implementation::get_nearest_charger(
   const rmf_traffic::agv::Planner::Start& start,
   const std::unordered_set<std::size_t>& charging_waypoints)
 {
-  assert(!charging_waypoints.empty());
+  if (charging_waypoints.empty())
+    return std::nullopt;
+
   const auto& graph = (*planner)->get_configuration().graph();
   Eigen::Vector2d p = graph.get_waypoint(start.waypoint()).get_location();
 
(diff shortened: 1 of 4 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -1009,10 +1010,18 @@ void FleetUpdateHandle::add_robot(
          fleet = shared_from_this()](
         rmf_traffic::schedule::Participant participant)
   {
-    const std::size_t charger_wp = fleet->_pimpl->get_nearest_charger(
-        start[0], fleet->_pimpl->charging_waypoints);
+    const auto charger_wp = fleet->_pimpl->get_nearest_charger(start[0]);
+    
+    if (!charger_wp.has_value())
+    {
+      throw std::runtime_error(
```
Yadunund, line 1017: Instead of throwing an error here, let me know if it is better to pass in a random waypoint as the charging waypoint but then warn users to call `RobotUpdateHandle::set_charger_waypoint()`.

mxgrey: If we get `nullopt` for the nearest charger, then that implies there are no chargers in the whole graph. That is almost certainly a system integration oversight and it makes sense to throw an exception for it.

rmf_fleet_adapter/src/rmf_fleet_adapter/phases/WaitForCharge.cpp
```diff
@@ -59,15 +59,23 @@ const std::string& WaitForCharge::Active::description() const
...
-  double charge_to_soc)
+  double charge_to_soc,
+  rmf_traffic::Time start_time)
 : _context(std::move(context)),
   _battery_system(battery_system),
-  _charge_to_soc(charge_to_soc)
+  _charge_to_soc(charge_to_soc),
+  _start_time(start_time)
 {
+
```
mxgrey, line 69: Nitpick: unnecessary newline

mxgrey (approved):
This looks good. There are just some very minor formatting nitpicks.

Also there's a [TODO here](https://github.com/open-rmf/rmf_ros2/blob/9c533ccb764b9068511a3b2e4c35b71b89058c66/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L757-L760) that we could consider addressing now. The planner performance should be good enough at this point that we can quickly and easily generate a plan to each charger waypoint to see which one will actually be the closest.

Yadunund:
Thanks for the quick review @mxgrey! I have addressed the TODO. To determine the nearest charger, I find the charging waypoint with the lowest `ideal_cost` as reported by the Result generated by the planner.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -738,28 +738,22 @@ auto FleetUpdateHandle::Implementation::is_valid_assignments(
...
+  std::optional<std::size_t> nearest_charger = std::nullopt;
   for (const auto& wp : charging_waypoints)
   {
-    const auto loc = graph.get_waypoint(wp).get_location();
-    // TODO: Replace this with a planner call
-    // when the performance improvements are finished
-    const double dist = (loc - p).norm();
-    if (dist < min_dist)
+    const rmf_traffic::agv::Planner::Goal goal{wp};
+    const auto& planner_result = (*planner)->plan(start, goal);
```
mxgrey, line 752: Since you're only using `ideal_cost` and not the actual plan that gets generated, you can use [`setup`](https://github.com/open-rmf/rmf_traffic/blob/f93fc723c39d8112f601877b3422d63296a0aed2/rmf_traffic/include/rmf_traffic/agv/Planner.hpp#L497-L500) here instead of `plan`. That will also give back the ideal cost but it won't require any planner iterations, so it should always be faster.

Yadunund: Thanks for the suggestion!

mxgrey (approved):
The code looks good, and I ran a few simulations which all ran with no problem :+1:

PR #48, opened 2021-05-03
Improve arrival estimates to account for holding times

mxgrey:
The fleet adapter API has a misleading name for one of its function signatures: `ArrivalEstimator`. This name implies that it expects the user to estimate when they will arrive at a waypoint, but if the estimator is given a time that would have the robot arrive ahead of its original schedule, then the fleet adapter implementation will push the whole schedule forward, even though the robot should actually stop and wait at that point. This creates more scheduling shifting than what we should really see.

This PR changes the implementation for `ArrivalEstimator` so that if an input would put the robot ahead of its originally planned schedule, then the time shift will be truncated to assume that the robot will intentionally stop and wait at the waypoint until the waypoint's `time()` field is reached.


Yadunund (approved):
Tested and works great!

PR #50, opened 2021-05-03
Correct version for rmf_task_ros2

Yadunund:


mxgrey:
Due to build farm (or bloom?) limitations, all of the packages within a single repo need to keep the same version, so we'll have to close this PR.

This does raise the question of whether we should move `rmf_task_ros2` into its own repo.

PR #53, opened 2021-05-07
Publish phase info in TaskSummary

youliangtan:
## Feature
Previously we had a discussion on having a better way to represent phases for each robot. According to the `rmf-web` wireframe [here](https://www.figma.com/proto/nFiXHduImJev6xTodDpd0K/TNG-WIP?node-id=3%3A5816&scaling=min-zoom), visualizing phases is important to inform the user about the task status, also for debugging. The current way of getting the phase is through the string in `TaskSummary.status` via regex. https://github.com/open-rmf/rmf-web/pull/305

Having a `Phase.msg` will make things simpler, while providing more info about the status of each phase. Phases are represented in multiple `tree_level`

### office World

- `rmf_internal_msgs`: branch `feature/phase_msg` https://github.com/open-rmf/rmf_internal_msgs/pull/14

Run `dispatch_loop` and `dispatch_delivery`, then observe the changes on `/task_summaries` topic

The printout is quite messy, thus a utility script `simple_task_summary_pub,py` is provided [here](https://gist.github.com/youliangtan/85449633eba66a73fd2496d3cb01a7e0), Run: 
```bash
python3 simple_task_summary_sub.py 
```

### Printout from helper script
```bash
# ======================================================
# Classic Office world Delivery Task
[tinyRobot2]  : move robot success | Remaining phases: 1 | Remaining phases: 2

 |__ Current: GoToPlace  	| Pending:   -> IngestItem
 				| Completed:   -> GoToPlace -> DispenseItem
	 |__ Current: MoveRobot  	| Pending:  
 					| Completed:   -> MoveRobot -> OpenDoor -> MoveRobot -> CloseDoor

# ======================================================
# Loop Task: num_loops = 3
[tinyRobot1]  : Moving [tinyRobot/tinyRobot1]: ( 10.0861 -6.97944  1.32862) -> ( 8.91186 -6.18135  2.76447) | Remaining phases: 5 | Remaining phases: 6

 |__ Current: GoToPlace  	| Pending:   -> GoToPlace -> GoToPlace -> GoToPlace -> GoToPlace -> GoToPlace
 				| Completed:  
	 |__ Current: MoveRobot  	| Pending:   -> OpenDoor -> MoveRobot -> CloseDoor -> MoveRobot
 					| Completed: 

======================================================
# Clinic world loop task (L1 to L2 via lift)
[deliveryRobot_2]  : [deliveryRobot_2] still waiting for lift [lift_1]  current state: L1 vs L2 | 1 vs 2 | deliveryRobot/deliveryRobot_2 vs deliveryRobot/deliveryRobot_2 | Remaining phases: 4 | Remaining phases: 1

 |__ Current: GoToPlace  	| Pending:  
 				| Completed:   -> GoToPlace
	 |__ Current: RequestLift  	| Pending:   -> MoveRobot -> EndLIftSession -> MoveRobot
 					| Completed:   -> MoveRobot -> RequestLift -> MoveRobot
```

Signed-off-by: youliang <tan_you_liang@hotmail.com>


tanyouliang95:
@koonpeng I think this will be helpful for the task api-server to gauge the progress of each task. Do let me know if this is informative enough.

mxgrey (changes requested):
Given the incoming changes to the `TaskSummary` message definition, we'll of course need to rework this PR in a number of ways.

koonpeng:
> @koonpeng I think this will be helpful for the task api-server to gauge the progress of each task. Do let me know if this is informative enough.

The current task progress does an naive assumption of the time it takes for a task to complete. Should we change the progress calculation from "prediction of how far along is the task" to "how much work has been done"? i.e. "progress = elapsed_time / predicted time" or "progress = work done / total work". 

The advantage of the former is that you will get a linear progress, which can accurately reflect how much of the task is done if the prediction is close, the disadvantage is that if prediction is wrong because of some delays, you get weird UX where the task seems to be progressing but is infact stuck, eventually it would hit the "99% problem" where a task is "stucked" at 99% and not done for a long time.

The "work done" would not suffer from the "fake progress" problem, a stucked task can be identified early, but the disadvantage is that you don't get linear or granular progress, it will jump from X% to Y% as each phase is done.

mxgrey:
If we want a really meaningful representation of "progress" we may want to get creative with our calculations. We could make this calculation arbitrarily complex, but I might suggest the following relatively simple approach to begin:

`(original_total_task_time_prediction - current_remaining_task_time_prediction)/original_total_task_time_prediction`

This carries the risk of "backtracking" if the robot gets rerouted down a longer route than what was originally predicted. There's even a risk of it taking on a negative value if the reroute is significant enough. Taking that into account, we should probably use `max(highest_progress, latest_progress_calculation)` to prevent backtracking, since it could cause confusion for observers. If we want the UI to express delay information, I think we could represent that separately from progress.

youliangtan:
close this, since moving towards to new "rmf_task 2.0" api.

PR #54, opened 2021-05-07
adding failover mode, not enabled by default (signed)

marcoag:
### Implemented feature

This feature adds failover mode to the `robot_state_aggregator` node. This mode is disabled by default. This mode needs the [stubborn_buddies](https://github.com/open-rmf/stubborn_buddies) library at compilation time and it has to be enabled at runtime through the `failover_mode` argument.

This feature uses the [stubborn_buddies](https://github.com/open-rmf/stubborn_buddies) libraries for the hearbeat and watchdog nodes. The linktime_composition.cpp is also heavily based on the one from `stubborn_buddies. 

This PR does not add a dependency to `stubborn_buddies`, failover mode will only be added if the library is found at compilation time, otherwise `robot_state_aggregator` will be compiled without failover mode. If `robot_state_aggregator` was compiled without failover mode and the user tries to select it at runtime the process will throw an error and exit.


arjo129 (approved):
Tested on #45

marcoag:
Just tested this branch on my side too

PR #60, opened 2021-05-27
Publish actives tasks for rmf-web

tanyouliang95:
This is to address: https://github.com/open-rmf/rmf-web/issues/352, specifically the  `Cleanup task entries in Tasks tab` (https://github.com/open-rmf/rmf-web/issues/352#issuecomment-847545871). 

Currently, the only way to get all task status from the dispatcher is via the `/get_tasks` ros service call. This is not helpful for high-volume continuous queries. 

Thus to solve this, the dispatcher will publish a list of aggregated active task_summaries to  `/dispatcher_active_tasks` topic. This will then be subscribed by the `rmf-web`'s `api-server`.

### DIFF
--- a/rmf_task_ros2/include/rmf_task_ros2/StandardNames.hpp
+++ b/rmf_task_ros2/include/rmf_task_ros2/StandardNames.hpp
@@ -33,7 +33,7 @@ const std::string GetTaskListSrvName = "get_tasks";
 const std::string TaskRequestTopicName = Prefix + "dispatch_request";
 const std::string TaskAckTopicName = Prefix + "dispatch_ack";
 const std::string TaskStatusTopicName = "task_summaries";
-const std::string ActiveTasksTopicName = "dispatch_active_tasks";
+const std::string ActiveTasksTopicName = "dispatcher_active_tasks";
 
 } // namespace rmf_task_ros2
 
--- a/rmf_task_ros2/include/rmf_task_ros2/StandardNames.hpp
+++ b/rmf_task_ros2/include/rmf_task_ros2/StandardNames.hpp
@@ -33,7 +33,7 @@ const std::string GetTaskListSrvName = "get_tasks";
 const std::string TaskRequestTopicName = Prefix + "dispatch_request";
 const std::string TaskAckTopicName = Prefix + "dispatch_ack";
 const std::string TaskStatusTopicName = "task_summaries";
-const std::string ActiveTasksTopicName = "dispatcher_active_tasks";
+const std::string ActiveTasksTopicName = "dispatcher_ongoing_tasks";
 
 } // namespace rmf_task_ros2
 
--- a/rmf_task_ros2/include/rmf_task_ros2/StandardNames.hpp
+++ b/rmf_task_ros2/include/rmf_task_ros2/StandardNames.hpp
@@ -33,6 +33,7 @@ const std::string GetTaskListSrvName = "get_tasks";
 const std::string TaskRequestTopicName = Prefix + "dispatch_request";
 const std::string TaskAckTopicName = Prefix + "dispatch_ack";
 const std::string TaskStatusTopicName = "task_summaries";
+const std::string ActiveTasksTopicName = "dispatch_active_tasks";
 
 } // namespace rmf_task_ros2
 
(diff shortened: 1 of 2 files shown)


koonpeng:
1. What is considered an active task? Does it include tasks that are pending, queued, completed, failed etc?
2. After this is merged, is `/task_summaries` considered deprecated and should not be used anymore?

koonpeng (commented):
1. What is considered an active task? Does it include tasks that are pending, queued, completed, failed etc?
2. After this is merged, is `/task_summaries` considered deprecated and should not be used anymore?

tanyouliang95:
I think the word "active" starts to get a little bit ambiguous, we might need a better name for this. Let's call it `Ongoing` task for now:

- Ongoing Task:  with State: `Active`, `Pending`, `Queued`
- Terminated Task: with State: `Completed`, `Canceled`, `Failed`

The dispatcher will publish a list of on-going tasks to a topic, and the `api-server` will subscribe to it (let's name the topic as `dispatcher_ongoing_tasks` for now). Therefore, from the `api-server` point of view, it doesn't need to keep track of the state machine for each task. That being said, as `/task_summaries` is an "internal msgs" in rmf, the `api-server` will not need to subscribe to it.

Hopefully this explains the confusion. Let me know if this makes sense.

koonpeng:
What happens when a task goes from an ongoing state to a terminated state? Is it published once in `dispatcher_ongoing_tasks` and never published again?

> That being said, as /task_summaries is an "internal msgs" in rmf, the api-server will not need to subscribe to it.

Going off topic abit, is all the messages/topics in `rmf_internal_msgs` considered internal? In that case, do you know whats the recommended way to get all the information like door/lift/fleet states with a public topic?

mxgrey:
> Going off topic abit, is all the messages/topics in rmf_internal_msgs considered internal? In that case, do you know whats the recommended way to get all the information like door/lift/fleet states with a public topic?

The use of "internal" in `rmf_internal_msgs` is meant to signify that these messages are subject to change and should not be considered stable API. The web servers and clients that we develop don't really have a choice but to use these messages and rapidly adapt as we make changes to the message definitions.

Ideally for the sake of external users, we would develop web tools that have a stable API to encapsulate the `rmf_internal_msgs`. But that is not currently as important as rapidly prototyping the web functionality that we need.

Even more ideally it would be great to eventually stabilize the API+ABI of these message definitions, perhaps by taking advantage of OMG's DDS-XTypes (which I believe allows for an extensible wire protocol with stable API+ABI even as message definitions change), but I'm not currently clear on when/if/how ROS2 is going to support that.

tanyouliang95:
> What happens when a task goes from an ongoing state to a terminated state? Is it published once in dispatcher_ongoing_tasks and never published again?

Currently, if an "ongoing" task gets terminated, it will directly get popped out of the `ongoing` tasks list (won't publish once). As discussed earlier, we can definitely make it such that it will publish the terminated state once. 

Note that we will still retain `/get_tasks` srv, this will be helpful for us to get terminated tasks since they will remain unchanged most of the time. Now the debate is if the `dispatcher_node` should even store a recent history of terminated tasks, since moving forward, the api-server/db can easily achieve this.

koonpeng:
> Currently, if an "ongoing" task gets terminated, it will directly get popped out of the ongoing tasks list (won't publish once). As discussed earlier, we can definitely make it such that it will publish the terminated state once.

I think this will be important, else the server will never know when a task completes or is cancelled.

> Note that we will still retain /get_tasks srv, this will be helpful for us to get terminated tasks since they will remain unchanged most of the time.

Even though their status will be unchanged most of the time, the web server does not know when they will change, and so the only way is to continuously poll all tasks.

> Now the debate is if the dispatcher_node should even store a recent history of terminated tasks, since moving forward, the api-server/db can easily achieve this.

There is one scenario which the `/get_tasks` service will be useful, if the web server goes down, it will not know about task updates while it is offline, so on startup it uses the service to synchronize the tasks. Unlike other data like door and fleet states which will always be continuously published, a task that is completed will be removed from the list of published task, and so the web server will never know about these tasks that completed while it is offline.

tanyouliang95:
@koonpeng I have renamed the topic to `/dispatcher_ongoing_tasks`. and implemented the "publish terminated task" once behavior for this onging task topic. Let me know if it makes sense. Thanks.

koonpeng (approved):

PR #67, opened 2021-06-14
rewrite include to be packaging friendly

marcoag:
Signed-off-by: Marco A. Gutierrez <marco@openrobotics.org>

### Fixed bug

Include fails to find `.hpp` file when packaging:
https://build.ros2.org/view/Rbin_uF64/job/Rbin_uF64__rmf_traffic_ros2__ubuntu_focal_amd64__binary/3/console

### Fix applied

Rewrite include to work at the package level

### DIFF
--- a/rmf_traffic_ros2/src/update_participant/main.cpp
+++ b/rmf_traffic_ros2/src/update_participant/main.cpp
@@ -21,7 +21,7 @@
 #include <rclcpp/node.hpp>
 #include <rclcpp/executors.hpp>
 
-#include "../src/rmf_traffic_ros2/schedule/internal_YamlSerialization.hpp"
+#include "../rmf_traffic_ros2/schedule/internal_YamlSerialization.hpp"
 
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


rmf_traffic_ros2/src/update_participant/main.cpp
```diff
@@ -21,7 +21,7 @@
 #include <rclcpp/node.hpp>
 #include <rclcpp/executors.hpp>
 
-#include "../../rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/internal_YamlSerialization.hpp"
+#include "../src/rmf_traffic_ros2/schedule/internal_YamlSerialization.hpp"
```
mxgrey, line 24: I'm not quite sure why different environments are inconsistent about finding the header, but both of these paths look unreliable to me.

I believe we want this, to get the path of the header, relative to this `main.cpp` file:

```
#include "../rmf_traffic_ros2/schedule/internal_YamlSerialization.hpp"
```

marcoag: True, my path was wrong. For some reason the path seemed to be working.

Fixed in https://github.com/open-rmf/rmf_ros2/pull/67/commits/f8375ab9e7ffa3ffca5e0f591ed3beb30d901823

mxgrey (approved):

PR #70, opened 2021-06-22
exporting dependencies on rmf_traffic_ros2

marcoag:
`rmf_task_ros2` fails to find `yaml-cpp` [here](https://github.com/open-rmf/rmf_visualization/blob/main/rmf_visualization_schedule/CMakeLists.txt#L15), however the dependency is only found on the upstream `rmf_traffic_ros2`. I believe exporting the dependencies at the `rmf_traffic_ros2` `CMakeLists.txt` level should fix the issue.

### DIFF
--- a/rmf_traffic_ros2/CMakeLists.txt
+++ b/rmf_traffic_ros2/CMakeLists.txt
@@ -70,7 +70,15 @@ target_include_directories(rmf_traffic_ros2
 )
 
 ament_export_targets(rmf_traffic_ros2 HAS_LIBRARY_TARGET)
-ament_export_dependencies(rmf_traffic rmf_traffic_msgs rclcpp)
+ament_export_dependencies(
+  rclcpp
+  rmf_traffic
+  rmf_traffic_msgs
+  rmf_fleet_msgs
+  Eigen3
+  rclcpp
+  yaml-cpp
+)
 
 # TODO(MXG): Change these executables into shared libraries that can act as
 # ROS2 node components


rmf_traffic_ros2/CMakeLists.txt
```diff
@@ -70,7 +70,15 @@ target_include_directories(rmf_traffic_ros2
 )
 
 ament_export_targets(rmf_traffic_ros2 HAS_LIBRARY_TARGET)
-ament_export_dependencies(rmf_traffic rmf_traffic_msgs rclcpp)
+ament_export_dependencies(
+  rclcpp
+  rmf_traffic
+  rmf_traffic_msgs
+  rmf_fleet_msgs
+  Eigen3
```
mxgrey, line 78: I'm pretty sure `Eigen3` should be covered transitively by `rmf_traffic`. Is that not the case? In fact we shouldn't even need the `find_package(Eigen3 REQUIRED)` up above.

[Here's the relevant line in `rmf_traffic`.](https://github.com/open-rmf/rmf_traffic/blob/f9b5836e179f52309e9d27c4335196be133b5e04/rmf_traffic/cmake/rmf_traffic-config.cmake.in#L7)

mxgrey (approved):
Looks good, although if I put my pedantic hat on, I think we have an extra explicit dependency, as mentioned in my inline comment.

rmf_traffic_ros2/CMakeLists.txt
marcoag: True, it seems `find_package` could be removed (haven't tested it though), but even though, since it's also directly used by this package, shouldn't we keep it in the `CMakeLists.txt` for clarity and/or future refactors of this code?

mxgrey: Sure, I don't have any objection to keeping it :+1:

PR #75, opened 2021-07-10
Remove duplicate judgment on dropoff_ingestor property

0to1:
## Bug fix

if (delivery.dropoff_ingestor.empty())
    {
      RCLCPP_ERROR(
        node->get_logger(),
        "Required param [delivery.dropoff_ingestor] missing in TaskProfile."
        "Rejecting BidNotice with task_id:[%s]" , id.c_str());

      return;
    }

As in the above code, there are two same judgments in the code, and one needs to be removed.


mxgrey (approved):
Thanks for catching this!

mxgrey:
Ordinarily we require [GPG signing of commits](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/about-commit-signature-verification) to validate the identities of contributors and ensure that no bad actors are sneaking commits into our codebase.

Given the simplicity of this PR and how obvious of a change it is, we'll make an exception in this case, but for any more complex change we will require GPG signing on all commits.

0to1:
> Ordinarily we require [GPG signing of commits](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/about-commit-signature-verification) to validate the identities of contributors and ensure that no bad actors are sneaking commits into our codebase.
> 
> Given the simplicity of this PR and how obvious of a change it is, we'll make an exception in this case, but for any more complex change we will require GPG signing on all commits.

PR #88, opened 2021-07-27
remove self-generated task types in dispatcher during reassignment

youliangtan:
Previously, dispatcher_node could only cancel tasks `ChargeBattery` task during reassignment, since `ChargeBattery` is auto-generated by RMF. Other task types which are self-generated (non-user submitted) will not get canceled during reassignment. This will solve this.

### DIFF
--- a/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
+++ b/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
@@ -61,7 +61,7 @@ public:
   /// of with task STATE_ACTIVE
   DispatchTasks active_dispatch_tasks;
   DispatchTasks terminal_dispatch_tasks;
-  DispatchTasks user_submitted_tasks;  // store ongoing user submitted tasks
+  std::set<std::string> user_submitted_tasks;  // ongoing submitted task_ids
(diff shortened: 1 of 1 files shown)


Yadunund:
Hello, build failed with this errror:
```
   <testcase classname="test_rmf_fleet_adapter.MockAdapterFixture" name="Scenario: ingest item phase" time="0.0">
      <error type="TEST_CASE">
SIGSEGV - Segmentation violation signal
at /home/runner/work/rmf_ros2/rmf_ros2/ros_ws/src/bd883j35yy/rmf_ros2/rmf_fleet_adapter/test/phases/test_IngestItem.cpp:53
      </error>
    </testcase>
```
at this line: https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/test/phases/test_IngestItem.cpp#L53

Test results can be downloaded here ( at the bottom ): 
https://github.com/open-rmf/rmf_ros2/actions/runs/1069793498

youliangtan:
After "rerun" all the checks, it seems to pass. :thinking: It is also quite weird that the previously `test_IngestItem` failed, considering it's not relevant.

Yadunund (approved):

PR #95, opened 2021-08-06
check if lane request's fleet_name is equal to the fleet's fleet_name

lkw303:
### Fixed bug
The Fleet Adapter publishes closed lanes even though the LaneRequest messages were not directed at them. Tjis causes all the fleet adapters to publish a ClosedLanes Message even though the LaneRequest was not meant for them.

### Fix applied
Added an if statement to check if fleet_name in  LaneRequest is equal to the fleet_name of the fleet.


mxgrey (dismissed):
Thanks for catching this! It looks good to me.

One thing to consider: if the `request_msg->fleet_name` is blank, then should we apply the request to all fleets? Otherwise to close a lane for all fleets, a user would have to issue a separate lane request to each of them.

mxgrey:
By the way, we have a "DCO" requirement (helps us avoid lawsuits from patent trolls) which your commit is not currently satisfying. Before we can merge this PR, we'll need you to follow the instructions here: https://github.com/open-rmf/rmf_ros2/pull/95/checks?check_run_id=3259667077

lkw303:
> One thing to consider: if the `request_msg->fleet_name` is blank, then should we apply the request to all fleets? Otherwise to close a lane for all fleets, a user would have to issue a separate lane request to each of them.

Hi, sorry,  Can I clarify shall I make another commit to add this in?

mxgrey:
> Hi, sorry, Can I clarify, shall I make another commit to add this in?

It looks like you've followed the instructions correctly, so now the DCO check is approved. Thank you!

mxgrey (dismissed):

mxgrey:
> Hi, sorry, Can I clarify, shall I make another commit to add this in?

Oh sorry, I misread the original question here.

I think it's fine either way. If you agree with my suggestion then feel free to create add a commit for it. Otherwise we can leave it as-is and potentially add in the behavior I recommended at another time if anyone requests it.

mxgrey:
Oh, I apologize for how nit-picky this is, but the code style checker is complaining about trailing whitespace on lines 826 and 828.

Line 826 has 8 trailing spaces while line 828 has 7 trailing spaces i.e.

```
      if (!connections)
        return;
        ⏎
      if (request_msg->fleet_name != fleet_name)
        return;       ⏎
```

The code style checker won't let the PR pass until those trailing spaces are removed.

mxgrey:
In case you're using Visual Studio Code I strongly recommend the [Strict Whitespace](https://marketplace.visualstudio.com/items?itemName=sidp.strict-whitespace) plugin, as well as these settings in your `settings.json`:

```
"files.trimTrailingWhitespace": true,
"files.insertFinalNewline": true
```

lkw303:
Oh!  I see sorry.. my bad. I also added an additional check to see if the fleet_name is empty which you mention previously. Additionally, I made changes to the close_lane node as well. Otherwise if the fleet_name is empty the request would never be fulfilled as the fleet_name from the request will never match the fleet_name of the ClosedLanes messages.

rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -824,6 +824,9 @@ std::shared_ptr<Connections> make_fleet(
       if (!connections)
         return;
 
+      if (request_msg->fleet_name != fleet_name && !request_msg->fleet_name.empty())
```
mxgrey, line 827: Sorry again for how nitpicky the style checker is, but this line seems to violate the 80-character limit. The style checker requires this to be changed to

```
      if (request_msg->fleet_name != fleet_name &&
      !request_msg->fleet_name.empty())
```

I think once that's fixed we'll be able to merge this. Thanks for bearing with our CI process :bow:

mxgrey (changes requested):

mxgrey (dismissed):

lkw303:
Hi, sorry but I just noticed that the merging is block as it says that the base branch requires all commits to be signed. Can I check if I should I do a git rebase to sign all my previous commits on my branch?

mxgrey:
Right, the easiest way I know of for fixing that is to set up the GPG signing using [these instructions](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/about-commit-signature-verification) and then doing a rebase.

Apologies again for all these contribution barriers. Luckily once the GPG signing is set up once, it will work forever (per computer that you use).

lkw303:
Hi! Just merged with main and have signed off and verified all my previous commits. Hopefully its all good now. Sorry for all the trouble.

PR #108, opened 2021-08-18
Accommodate finishing task

Yadunund:
PR https://github.com/open-rmf/rmf_task/pull/28 allows for a `RequestFactory` to be supplied when task allocation planning. The request generated by the `RequestFactory` is automatically appended to the end of the task queue for all agents. This is very useful when you want robots in a fleet to always perform a certain task after finishing other requested tasks. For example, if we do not want robots to remain idle after their last performed task and instead want them to recharge their batteries, we can supply the `ChargeBatteryFactory` to the task planner which will automatically append a `ChargeBattery` task to the end of the robot's task queues.

This PR:
* Update `FleetUpdateHandle::set_task_planner_params()` to optionally accept a `RequestFactory`
* Updates `full_control` launch file and src to allow users to set the fleet adapter's task planner with either `ChargeBatteryFactory` or `ParkRobotFactory`

### DIFF
--- a/rmf_fleet_adapter/launch/fleet_adapter.launch.xml
+++ b/rmf_fleet_adapter/launch/fleet_adapter.launch.xml
@@ -16,7 +16,6 @@
   <arg name="angular_acceleration" description="The nominal angular acceleration of the vehicles in this fleet"/>
   <arg name="footprint_radius" description="The radius of the footprint of the vehicles in this fleet"/>
   <arg name="vicinity_radius" description="The radius of the personal vicinity of the vehicles in this fleet"/>
-  <arg name="perform_deliveries" default="false" description="Whether this fleet adapter can perform deliveries"/>
   <arg name="delay_threshold" default="10.0" description="How long to wait before replanning" />
   <arg name="disable_delay_threshold" default="false" description="Disable the delay_threshold behavior" />
   <arg name="retry_wait" default="10.0" description="How long a retry should wait before starting"/>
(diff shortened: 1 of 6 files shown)


mxgrey (dismissed):
These changes look good to me, although I think it's highlighting the need to switch to dynamically defined tasks instead of these static task types.

We just need to wait for https://github.com/open-rmf/rmf_task/pull/28 before merging.

rmf_fleet_adapter/launch/fleet_adapter.launch.xml
```diff
@@ -27,7 +27,7 @@
   <arg name="perform_deliveries" default="false" description="Whether this fleet adapter can perform deliveries"/>
   <arg name="perform_cleaning" default="false" description="Whether this fleet adapter can perform cleaning"/>
   <arg name="finishing_request_charge_battery" default="false" description="Whether robots in this fleet should perform a ChargeBattery task once all queued tasks are completed"/>
-  <arg name="finishing_request_return_charger" default="false" description="Whether robots in this fleet should return to their chargers once all queued tasks are completed"/>
+  <arg name="finishing_request_park_robot" default="false" description="Whether robots in this fleet should return to their parking spots once all queued tasks are completed"/>
```
mxgrey, line 30: Sorry for just thinking of this now, but I wonder if we should refactor these finishing request parameters to just one:

```
<arg name="finishing_request" default="nothing" description="What should happen when the robot is finished with its tasks? Options: nothing (default), park, charge"/>
```

As it is now, it's kind of confusing what might happen if `finishing_request_charge_battery` and `finishing_request_park_robot` are both `true`.

mxgrey: If a user passes an unrecognized string, we can just print out a warning and default to `nothing` behavior.

Yadunund: Fair point. But how would users know what the valid strings are for different factories? It seems that the mapping will be an implementation detail.

mxgrey: Right, that's why I recommend putting all the supported strings into the parameter description (see the `Options: ...` at the end of the code snippet that I recommended).

Off the top of my head, I don't know how visible that parameter description will be to end-users. But we could also print out the list of supported parameters in the warning that we issue when we receive an invalid string.

Yadunund: Roger. I've made the requested change!

mxgrey (approved):
Looks great!

I'm going to merge https://github.com/open-rmf/rmf_task/pull/28 first and then watch CI on this PR to make sure it's all good before merging.

PR #110, opened 2021-08-31
Add read_only_blockade adapter

mxgrey:
This PR introduces the `read_only_blockade` fleet adapter which is an improved version of the old `read_only` adapter. The API for integrating with `read_only_blockade` is different from the API for integrating with `read_only` so I'm keeping them as separate adapters for the sake of backwards compatibility.

This PR also includes a few tweaks that made performance and behavior more reliable for DP3.

### DIFF
--- a/rmf_fleet_adapter/src/read_only_blockade/main.cpp
+++ b/rmf_fleet_adapter/src/read_only_blockade/main.cpp
@@ -17,11 +17,7 @@
 
 #include "FleetAdapterNode.hpp"
 
-#include <rmf_traffic/geometry/Circle.hpp>
-
 #include <rclcpp/executors.hpp>
-
-#include "../rmf_fleet_adapter/ParseArgs.hpp"
 
 using namespace rmf_fleet_adapter;
 
--- a/rmf_fleet_adapter/src/read_only_blockade/main.cpp
+++ b/rmf_fleet_adapter/src/read_only_blockade/main.cpp
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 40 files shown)


rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp
```diff
@@ -124,6 +125,26 @@ class Negotiation
     rmf_traffic::schedule::ParticipantId for_participant,
     std::unique_ptr<rmf_traffic::schedule::Negotiator> negotiator);
 
+  std::shared_ptr<void> register_negotiator(
```
Yadunund, line 128: Should we add some documentation for this overload? I think the parameter descriptions for the overload below need to be updated too?

rmf_fleet_adapter/src/read_only_blockade/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#include "FleetAdapterNode.hpp"
+
+#include <rmf_traffic/geometry/Circle.hpp>
```
Yadunund, line 20: This include does not seem to be used.

rmf_fleet_adapter/src/read_only_blockade/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <rclcpp/executors.hpp>
+
+#include "../rmf_fleet_adapter/ParseArgs.hpp"
```
Yadunund, line 24: This include does not seem to be used.

Yadunund (commented):
Left a couple of minor comments.
Should we also update the descriptions in the launch file to let users know of the `read_only_blocade` node? 
https://github.com/open-rmf/rmf_ros2/blob/9078effb21f3eb6e9ba673257692be86d649ac34/rmf_fleet_adapter/launch/fleet_adapter.launch.xml#L7-L12

rmf_fleet_adapter/src/read_only_blockade/main.cpp
mxgrey: 52488fb

mxgrey:
> Should we also update the descriptions in the launch file to let users know of the read_only_blocade node?

Great point! Added: 68b8bcd

mxgrey:
Good catch! Fixed: 06c0faa

Yadunund (approved):

Yadunund:
`uncrustify` failing 💀

Yadunund (approved):

PR #111, opened 2021-09-01
Pybind/finishing task

Yadunund:
* Update bindings with changes from https://github.com/open-rmf/rmf_ros2/pull/108
* Add binding for `RobotUpdateHandle::maximum_delay(value)`

Note: The first change affects ABI stability so we should merge this in before releasing into `galactic`

### DIFF
--- a/rmf_fleet_adapter_python/src/adapter.cpp
+++ b/rmf_fleet_adapter_python/src/adapter.cpp
@@ -187,12 +187,12 @@ PYBIND11_MODULE(rmf_adapter, m) {
       if (finishing_request_string == "charge")
       {
         finishing_request =
-          std::make_shared<rmf_task::requests::ChargeBatteryFactory>();
+        std::make_shared<rmf_task::requests::ChargeBatteryFactory>();
       }
       else if (finishing_request_string == "park")
       {
         finishing_request =
-          std::make_shared<rmf_task::requests::ParkRobotFactory>();
+        std::make_shared<rmf_task::requests::ParkRobotFactory>();
       }
       else
       {
--- a/rmf_fleet_adapter_python/src/adapter.cpp
+++ b/rmf_fleet_adapter_python/src/adapter.cpp
@@ -18,6 +18,9 @@
 #include <rmf_battery/agv/SimpleMotionPowerSink.hpp>
 #include <rmf_battery/agv/SimpleDevicePowerSink.hpp>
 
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter_python/src/adapter.cpp
```diff
@@ -166,27 +179,47 @@ PYBIND11_MODULE(rmf_adapter, m) {
     battery::SimpleDevicePowerSink& t_sink,
     double recharge_threshold,
     double recharge_soc,
-    bool account_for_battery_drain)
+    bool account_for_battery_drain,
```
youliangtan, line 182: Just a comment, not sure should we have a default arg for this. (purely for backward compatibility for others using this api).

rmf_fleet_adapter_python/src/adapter.cpp
```diff
@@ -166,27 +179,47 @@ PYBIND11_MODULE(rmf_adapter, m) {
...
+    const std::string& finishing_request_string = "nothing")
```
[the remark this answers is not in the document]
Yadunund: Just matching the behaviour of the `full_control` launch file
https://github.com/open-rmf/rmf_ros2/pull/108#discussion_r698984681

rmf_fleet_adapter_python/src/adapter.cpp
Yadunund: default arg for `account_for_battery_drain`? That parameter has been there since the previous release. The newly added `finishing_request_string` takes in a default value to make the function call backwards compatible

youliangtan: My mistake, overlooked the diff

youliangtan (dismissed):

rmf_fleet_adapter_python/src/adapter.cpp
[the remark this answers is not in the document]
youliangtan: Okay, maybe for the new revamp (when more finishing behaviors are introduced), we can use similar to this https://github.com/open-rmf/rmf_ros2/blob/9078effb21f3eb6e9ba673257692be86d649ac34/rmf_fleet_adapter_python/src/adapter.cpp#L264

Yadunund:
@youliangtan need a re-review

youliangtan (approved):

PR #112, opened 2021-09-01
Eliminate remediation cycles

mxgrey:
It's been found that for large numbers of participants and/or mirrors, update messages may get dropped which creates harmful cycles of remedial update requests. This PR fixes this problems in two ways:

1. We enforce consistency about the versions that get bundled into updates. In the previous implementation, we allowed remedial update requests to potentially derail the baseline version of the updates that were getting issued. This could create issues where mirrors which were following along correctly end up needing their own remedial updates because suddenly an update that targets a new version has a baseline version that they weren't prepared for. With this PR, we no longer allow the baseline of mirror updates to be changed; instead we simply issue extra remedial updates (upon request) alongside the very consistent mainline updates.
2. We explicitly indicate which updates are remedial so that mirrors don't get confused if the baseline version of the update does not match their own.

This PR depends on this update to the traffic message definitions: https://github.com/open-rmf/rmf_internal_msgs/pull/24

### DIFF
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp
@@ -28,6 +28,8 @@
 #include <rmf_battery/DevicePowerSink.hpp>
 #include <rmf_battery/MotionPowerSink.hpp>
 
+#include <rmf_task/RequestFactory.hpp>
+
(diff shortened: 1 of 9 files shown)


Yadunund (approved):

Yadunund:
Looks like `uncrustify` is failing

mxgrey:
@Yadunund The commits to fix uncrustify will require a new approval.

Yadunund (approved):

Issue #117, opened 2021-09-10
Add support for nav graph initializaton and runtime updates through ROS2 topic


luca-della-vedova:
### Description

It would be great for applications that require maps to be update-able at runtime by adding the capability to fleet adapters to subscribe to a `/map` topic, like the one published by `building_map_server` in [rmf_building_map_tools](https://github.com/open-rmf/rmf_traffic_editor/blob/main/rmf_building_map_tools/building_map_server/building_map_server.py) that is used by other parts of the stack, for example `rmf_simulation` slotcar to initialize its position, or visualization nodes to know the layout of the building.

### Implementation considerations

We will need to add convert function between `rmf_traffic::agv::Graph` and `building_map_msgs::Graph`, as well as add a subscription to the `map` topic on the fleet adapter, with callbacks that update the navigation graph.
We will need to figure out whether we want to keep the `nav_graph` [file initialization in downstream nodes](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos/launch/include/adapters/tinyRobot_adapter.launch.xml#L17) or if we want to drop it altogether and make fleet adapters depending on the map topic being published.

There will be tricky edge cases such as "what happens if a map is updated while a robot is en route and waypoints suddenly change positions or are non existent anymore?".
There are different degrees of route invalidation that could be done such as:

- Invalidate all routes instantly, all tasks including currently executing ones are replanned.
- Allow finishing currently executing tasks on old map but replan all scheduled tasks.
- Allow finishing running and scheduled tasks on old map, only apply map update to newly submitted tasks.

I'm not sure which way is the best, depending on when trajectory generation is done at the `rmf_traffic` level either the third or both second and third option would be the easiest to implement but we could offer an API if we wanted to offer maximum flexibility, something along the lines of:

```
enum class MapUpdateBehavior {
  PreemptAll,
  PreemptNotRunning,
  NoPreempt
};
```

### Additional information

As a measure of the current level of integration of the `/map` topic in the RMF stack, currently in `office` demo, there are 5 subscribers to the `/map`, two simulation plugins (one per robot) and three visualization plugins:
```
$ros2 topic info -v /map
Type: rmf_building_map_msgs/msg/BuildingMap

Publisher count: 0

Subscription count: 5

Node name: building_systems_visualizer
Node namespace: /
Topic type: rmf_building_map_msgs/msg/BuildingMap
Endpoint type: SUBSCRIPTION
[...]

PR #120, opened 2021-09-14
Add graph conversion function from rmf_building_map_msgs to rmf_traffic

luca-della-vedova:
### Implemented feature

This PR is to work towards #117, it adds a function in `rmf_traffic_ros2` to convert from the [Graph](https://github.com/open-rmf/rmf_building_map_msgs/blob/main/rmf_building_map_msgs/msg/Graph.msg) implementation in `rmf_building_map_msgs` to the [Graph](https://github.com/open-rmf/rmf_traffic/blob/main/rmf_traffic/include/rmf_traffic/agv/Graph.hpp) in `rmf_traffic::agv`

The implementation is simple for now and it only implements waypoints without infrastructure.
Also, unlike the [implementation in parse_graph](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/parse_graph.cpp#L26-L29) it doesn't implement `orientation_constraints`. I couldn't find any place where those are used in the demos and it is unclear in the case of pure conversion between message formats where the `VehicleTraits` information would come from.
If we wanted support for doors / lifts it would be more complicated and probably not in the scope of `Graph` conversion since those are contained in the top level Building message and not in the graph itself.

In the scope of #117, the intended use for this PR is to allow users to subscribe to maps, select the graph that they want to use for their fleet, then use the newly introduced `convert` function to convert it to a `rmf_traffic::agv::Graph` that can be assigned to a planner.

I also added unit tests to verify that all the functionality works as intended (which helped me catch a sneaky `auto` instead of `auto&`. All the parts that are currently unimplemented are documented in the code but, as mentioned above, they might not be in the scope of a Graph conversion function.

### DIFF
--- a/rmf_traffic_ros2/CMakeLists.txt
+++ b/rmf_traffic_ros2/CMakeLists.txt
@@ -16,6 +16,7 @@ include(GNUInstallDirs)
 find_package(ament_cmake REQUIRED)
 find_package(rmf_traffic REQUIRED)
 find_package(rmf_traffic_msgs REQUIRED)
+find_package(rmf_building_map_msgs REQUIRED)
 find_package(rmf_fleet_msgs REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(rclcpp REQUIRED)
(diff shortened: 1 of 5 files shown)


luca-della-vedova:
There is ongoing discussion on changing the format that we use to store maps / graphs that would make this PR obsolete, converting to draft for the time being.

PR #128, opened 2021-09-30
Refactor tests - remove shared pointers from lambda captures

arjo129:
There are many places in our tests (and a few places in the `fleet_update_handle`) where we use the following anti-pattern:
```
auto some_var = std::make_shared<SomeType>();

auto lambda = [some_var]()
{
 ...
}

execute_asyncronously(lambda);
```

This is yet another way to trigger memory leaks due to the creation of a circular reference. In most places I have refactored this using `std::weak_ptr`s. In some places I've been forced to use `std::move()` on the `std::shared_ptr`.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -1007,6 +1007,9 @@ void FleetUpdateHandle::add_robot(
       rmf_traffic::schedule::Participant participant)
     {
       auto fleet = fleet_wptr.lock();
+      if (!fleet)
+        return;
+
       const auto charger_wp = fleet->_pimpl->get_nearest_charger(start[0]);
 
       if (!charger_wp.has_value())
@@ -1047,10 +1050,12 @@ void FleetUpdateHandle::add_robot(
         handle_cb = std::move(handle_cb)](const auto&)
         {
           auto fleet = fleet_wptr.lock();
-          if(!fleet) return;
+          if (!fleet)
+            return;
 
           auto node = node_wptr.lock();
-          if(!node) return;
+          if (!node)
+            return;
 
           
           // TODO(MXG): We need to perform this test because we do not currently
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -1044,7 +1044,7 @@ void FleetUpdateHandle::add_robot(
       // We schedule the following operations on the worker to make sure we do not
       // have a multiple read/write race condition on the FleetUpdateHandle.
       worker.schedule(
-        [fleet_wptr = std::weak_ptr<FleetUpdateHandle>(fleet), 
+        [fleet_wptr = std::weak_ptr<FleetUpdateHandle>(fleet),
         node_wptr = std::weak_ptr<Node>(fleet->_pimpl->node),
         context = std::move(context),
         handle_cb = std::move(handle_cb)](const auto&)
@@ -1057,7 +1057,7 @@ void FleetUpdateHandle::add_robot(
           if (!node)
             return;
 
-          
+
           // TODO(MXG): We need to perform this test because we do not currently
           // support the distributed negotiation in unit test environments. We
           // should create an abstract NegotiationRoom interface in rmf_traffic and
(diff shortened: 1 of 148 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -1003,9 +1003,10 @@ void FleetUpdateHandle::add_robot(
     command = std::move(command),
     start = std::move(start),
     handle_cb = std::move(handle_cb),
-    fleet = shared_from_this()](
+    fleet_wptr = weak_from_this()](
       rmf_traffic::schedule::Participant participant)
     {
+      auto fleet = fleet_wptr.lock();
```
mxgrey, line 1009: Whenever we lock a `std::weak_ptr`, we always need to check if the locking was successful and simply return right away if it wasn't, i.e.:

```
if (!fleet)
  return;
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -1032,14 +1035,24 @@ void FleetUpdateHandle::add_robot(
...
       // have a multiple read/write race condition on the FleetUpdateHandle.
       worker.schedule(
-        [context, fleet, node = fleet->_pimpl->node,
+        [fleet_wptr = std::weak_ptr<FleetUpdateHandle>(fleet), 
+        node_wptr = std::weak_ptr<Node>(fleet->_pimpl->node),
+        context = std::move(context),
         handle_cb = std::move(handle_cb)](const auto&)
         {
+          auto fleet = fleet_wptr.lock();
+          if(!fleet) return;
```
mxgrey, line 1050: There are some style tweaks needed here, i.e. a space in `if (` and a newline before `return;`.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -1032,14 +1035,24 @@ void FleetUpdateHandle::add_robot(
...
+
+          auto node = node_wptr.lock();
+          if(!node) return;
```
mxgrey, line 1053: Same here

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/test/MockAdapter.cpp
```diff
@@ -70,13 +70,15 @@ class MockScheduleNode : public rmf_traffic::schedule::Writer
   {
     _worker.schedule(
       [
-        database = _database,
+        database = std::weak_ptr<rmf_traffic::schedule::Database>(_database),
```
mxgrey, line 73: I would argue that the original use of `std::shared_ptr` here is fine. It just means the database will stay alive until the worker gets around to performing this operation. I don't believe there's any circular dependency issue here.

Same for the other database operations.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
arjo129: 73da754

mxgrey:
I'm seeing numerous style complaints from uncrustify. You could probably do searches for `if(` and `) return` to catch them all.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/test/MockAdapter.cpp
arjo129: You're right. I didn't notice that worker was owned by `MockAdapter` :man_facepalming: 703a3b5

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
arjo129: Yea I'll run uncrustify as a batch process.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/test/MockAdapter.cpp
mxgrey: My understanding of "workers" in RxCpp is that they're kind-of-sort-of global objects, so they aren't really "owned" in the usual sense. Whatever gets scheduled to them will go into a globally stored pool of scheduled events, and get executed eventually. Once the scheduled event is executed, the lambda will get released, and everything in its capture list gets destructed. So we shouldn't really have to worry about circularity when it comes to tasks that are scheduled to workers.

That being said, I may have some misunderstandings on this since RxCpp is quite complex, so if anyone knows better please do correct me on this.

mxgrey (approved):
Thanks for hunting down all these lingering issues!

PR #132, opened 2021-10-13
Add speed limit publishing to full control fleet adapter

luca-della-vedova:
### Implemented feature

Requires https://github.com/open-rmf/rmf_internal_msgs/pull/28 Add publishing of lane level speed limit to full_control adapter.

This PR parses the graph and publishes the speed_limit if one was found when reading the navgraph so the downstream simulated robots can respect it.

### DIFF
--- a/rmf_fleet_adapter/src/full_control/main.cpp
+++ b/rmf_fleet_adapter/src/full_control/main.cpp
@@ -214,7 +214,7 @@ public:
       location.yaw = p.z();
 
       // The speed limit is set as the minimum of all the approach lanes' limits
-      std::optional<double> speed_limit;
+      std::optional<double> speed_limit = std::nullopt;
       for (const auto& lane_idx : wp.approach_lanes())
       {
         const auto& lane = _travel_info.graph->get_lane(lane_idx);
@@ -227,7 +227,15 @@ public:
             speed_limit = lane_limit.value();
         }
       }
-      location.approach_speed = speed_limit.has_value() ? speed_limit.value() : 0.0;
+      if (speed_limit.has_value())
+      {
+        location.obey_approach_speed_limit = true;
+        location.approach_speed_limit = speed_limit.value();
+      }
+      else
+      {
+        location.obey_approach_speed_limit = false;
+      }
 
       // Note: if the waypoint is not on a graph index, then we'll just leave
       // the level_name blank. That information isn't likely to get used by the
--- a/rmf_fleet_adapter/src/full_control/main.cpp
+++ b/rmf_fleet_adapter/src/full_control/main.cpp
@@ -227,6 +227,7 @@ public:
             speed_limit = lane_limit.value();
         }
       }
+      location.approach_speed = speed_limit.has_value() ? speed_limit.value() : 0.0;
 
       // Note: if the waypoint is not on a graph index, then we'll just leave
       // the level_name blank. That information isn't likely to get used by the
--- a/rmf_fleet_adapter/src/full_control/main.cpp
+++ b/rmf_fleet_adapter/src/full_control/main.cpp
@@ -213,6 +213,21 @@ public:
       location.y = p.y();
       location.yaw = p.z();
 
+      // The speed limit is set as the minimum of all the approach lanes' limits
+      std::optional<double> speed_limit;
+      for (const auto& lane_idx : wp.approach_lanes())
+      {
+        const auto& lane = _travel_info.graph->get_lane(lane_idx);
+        const auto& lane_limit = lane.properties().speed_limit();
+        if (lane_limit.has_value())
+        {
+          if (speed_limit.has_value())
+            speed_limit = std::min(speed_limit.value(), lane_limit.value());
+          else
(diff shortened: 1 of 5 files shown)


rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -213,6 +213,22 @@ class FleetDriverRobotCommandHandle
...
+        const auto& lane_limit = lane.properties().speed_limit();
+        if (lane_limit.has_value())
+        {
+          if (speed_limit.has_value())
+            speed_limit = std::min(speed_limit.value(), lane_limit.value());
+          else
+            speed_limit = lane_limit.value();
+        }
+      }
+      location.approach_speed = speed_limit.has_value() ? speed_limit.value() : 0.0;
```
Yadunund, line 230: Wondering if we should use `NaN` or a negative number instead of `0.0`? speed = 0.0 does not intuitively convey that the robot should not respect any speed limits

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -814,6 +814,7 @@ rmf_fleet_msgs::msg::RobotState convert_state(const TaskManager& mgr)
         .x(p.x())
         .y(p.y())
         .yaw(l.orientation())
+        .approach_speed(0.0)
```
Yadunund, line 817: Populating the correct speed should be simple enough to include in this PR?

rmf_fleet_adapter/src/full_control/main.cpp
aaronchongth: I have similar concerns that I added on the review, https://github.com/open-rmf/rmf_internal_msgs/pull/28

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
luca-della-vedova: This is actually a bit tricky because the speed limit is currently calculated by knowing the speed limit of the [approach lanes](https://github.com/open-rmf/rmf_traffic/blob/main/rmf_traffic/include/rmf_traffic/agv/Planner.hpp#L738) which are a property of a `Plan` but not of the normal `Graph`, that this function uses to populate the state, at least as far as I understand

luca-della-vedova:
Update and tested against the [office world with a speed limit](https://gist.github.com/luca-della-vedova/9e81e864bbedd7a66bc9956cb7e2fe28) as shown in the [issue ticket](https://github.com/open-rmf/rmf_traffic/issues/43).
The main part I'm not too sure of is the mock_traffic_light implementation since I'm not that familiar with the codebase, it looks right but I didn't know how to test it and I'm happy to revert it to an unimplemented + TODO state if necessary.

rmf_fleet_adapter/src/mock_traffic_light/main.cpp
```diff
@@ -54,11 +54,14 @@ rmf_fleet_msgs::msg::Location make_location(
   rclcpp::Time t,
   const Eigen::Vector3d& p,
   const std::string& map_name,
-  const std::size_t index)
+  const std::size_t index,
+  const std::optional<double>& speed_limit)
 {
   return rmf_fleet_msgs::build<rmf_fleet_msgs::msg::Location>()
```
Yadunund, line 60: Maybe add a `const double approach_speed_limit = speed_limit.has_value() ? speed_limit.value() : 0.0;` which is used to assign the value below instead of `*speed_limit`

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -818,6 +818,8 @@ rmf_fleet_msgs::msg::RobotState convert_state(const TaskManager& mgr)
         .x(p.x())
         .y(p.y())
         .yaw(l.orientation())
+        .obey_approach_speed_limit(false)
```
Yadunund, line 821: Maybe we can create a ticket to correctly populate these fields for future work.

Yadunund (changes requested):
Looks good. Couple of minor comments

rmf_fleet_adapter/src/mock_traffic_light/main.cpp
luca-della-vedova: Done https://github.com/open-rmf/rmf_ros2/pull/132/commits/510dcf83de2fbf34d6919cad7f2f91e1b23d429f

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
luca-della-vedova: Done in #143

Yadunund (dismissed):

Yadunund (approved):

Issue #137, opened 2021-10-28
ASAN Reporting yet another Heap-use-after free.


arjo129:
### Required information:

- Operating system and version: Ubuntu
- OpenRMF installation type: Source
- OpenRMF version or commit hash: 0de31fc90512ad29b9349dada7f77b8a47847245
- ROS distribution and version: Galactic
- ROS installation type:
  - 
- Package or library, if applicable:
  - 

### Description of the bug
ASAN has detected a heap-use-after free in `rmf_fleet_adapter` full stack trace is here: [stacktrace.txt](https://github.com/open-rmf/rmf_ros2/files/7431851/stacktrace.txt)

### Steps to reproduce the bug

See #131 

### Expected behavior
No use-after free.

### Actual behavior
ASAN reports a use-after free.

Note: This is flaky behaviour and only occurs once in a blue moon.

mxgrey:
I think this might be fixed by https://github.com/open-rmf/rmf_ros2/pull/146

mxgrey:
This should be fixed by #146, so we'll close this issue. Feel free to reopen if it ever happens again.

PR #146, opened 2021-12-02
Safely capture the SearchForPath object

mxgrey:
There was a lingering `[this]` capture hiding in the source code. This PR fixes it by capturing a `weak_ptr` of the object and locking+checking it before usage.

This bug may explain https://github.com/open-rmf/rmf_ros2/issues/137


Yadunund (approved):
Looks good 😄

PR #147, opened 2021-12-20
increase publish fleet state frequency to 10hz

xiyuoh:
### Implemented feature

This PR updates the wall timer inside `fleet_state_publish_period` to 0.1 seconds to increase fleet state publish frequency for smoother visualization. 
This is to support the new demo fleet adapter PR open-rmf/rmf_demos#109.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -783,7 +783,7 @@ void FleetUpdateHandle::Implementation::fleet_state_publish_period(
   if (value.has_value())
   {
     fleet_state_timer = node->create_wall_timer(
-      std::chrono::seconds(1), [this]() { this->publish_fleet_state(); });
+      std::chrono::milliseconds(100), [this]() { this->publish_fleet_state(); });
   }
   else
   {
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -776,21 +776,6 @@ get_nearest_charger(
   return nearest_charger;
 }
 
-//==============================================================================
-void FleetUpdateHandle::Implementation::fleet_state_publish_period(
-  std::optional<rmf_traffic::Duration> value)
-{
-  if (value.has_value())
-  {
-    fleet_state_timer = node->create_wall_timer(
-      std::chrono::milliseconds(100), [this]() { this->publish_fleet_state(); });
-  }
-  else
-  {
-    fleet_state_timer = nullptr;
-  }
-}
-
 namespace {
 //==============================================================================
 rmf_fleet_msgs::msg::RobotState convert_state(const TaskManager& mgr)
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
@@ -208,7 +208,7 @@ public:
     handle->_pimpl->fleet_state_pub = handle->_pimpl->node->fleet_state();
     handle->_pimpl->fleet_state_timer =
       handle->_pimpl->node->try_create_wall_timer(
-      std::chrono::seconds(1), [me = handle->weak_from_this()]()
+      std::chrono::milliseconds(100), [me = handle->weak_from_this()]()
       {
         if (const auto self = me.lock())
           self->_pimpl->publish_fleet_state();


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -783,7 +783,7 @@ void FleetUpdateHandle::Implementation::fleet_state_publish_period(
   if (value.has_value())
   {
     fleet_state_timer = node->create_wall_timer(
-      std::chrono::seconds(1), [this]() { this->publish_fleet_state(); });
```
youliangtan, line 786: This function block `fleet_state_publish_period()` can be removed completely. since it is not used.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -208,7 +208,7 @@ class FleetUpdateHandle::Implementation
     handle->_pimpl->fleet_state_pub = handle->_pimpl->node->fleet_state();
     handle->_pimpl->fleet_state_timer =
       handle->_pimpl->node->try_create_wall_timer(
-      std::chrono::seconds(1), [me = handle->weak_from_this()]()
```
youliangtan, line 211: If the default value is changed, we will also need to update the comment [here] in the header. Hmmm also, not sure if we will need to change this 1s value, since the user can update it via the `fleet_state_publish_period` function.

youliangtan (changes requested):
Hi, thanks for identifying this. However, the `fleet_state_publish_period` function impl is already linked to [here](https://github.com/open-rmf/rmf_ros2/blob/ccf18ea463d979c4d23b8fba60df6812dd593b5e/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L1238-L1257). So feel free to update the pr to completely remove the function block.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
xiyuoh: Got it, done!

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
xiyuoh: Ah okay! For the demo adapter, the user just needs to modify the `robot_state_update_frequency` value ([example here](https://github.com/xiyuoh/rmf_demos/blob/feature/demo_fleet_adapter/rmf_demos/config/hotel/deliveryRobot_config.yaml#L52)), it will publish according to that frequency even if the default in `FleetUpdateHandle` is 1Hz. I initially thought we should increase the default frequency, but I can undo this if needed!

Yadunund: @xiyuoh the `robot_state_update_frequency` is not linked to the frequency of the published `FleetState` msgs. The former controls the rate at which the robots's position is updated to the `RobotUpdateHandle`. It does not publish any message. There is a separate timer within `FleetUpdateHandle` that retrieves the latest position/state of the robot and publishes a `FleetState` message. The periodicity of this timer is controlled by [this api](https://github.com/open-rmf/rmf_ros2/blob/ccf18ea463d979c4d23b8fba60df6812dd593b5e/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp#L179-L183). So ideally, both rates are set to the same value.

Yadunund:
I strongly recommend not to modify the implementation of this function. Instead, users should call [this api](https://github.com/open-rmf/rmf_ros2/blob/ccf18ea463d979c4d23b8fba60df6812dd593b5e/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp#L179-L183) to set the publish frequency to whatever they desire. 

Then in the python fleet adapter, users should use [this binding](https://github.com/open-rmf/rmf_ros2/blob/ccf18ea463d979c4d23b8fba60df6812dd593b5e/rmf_fleet_adapter_python/src/adapter.cpp#L234-L237) to set the publish frequency. Currently, the configuration is only checking if users want to publish FleetState and if so does not modify the default publish frequency as seen [here](https://github.com/open-rmf/fleet_adapter_template/blob/32f47a451ea67fa299a4fdb1a189d5db8df84b6b/fleet_adapter_template/fleet_adapter_template/fleet_adapter.py#L94-L95). What we can do instead is modify the config so that it takes in a frequency value instead of boolean. Then call ` fleet_handle.fleet_state_publish_period()` with the config value. The last thing to ensure is that the fleet state update frequency and robot state update frequency are the same.

If that makes sense, we can close this PR

xiyuoh:
I see! Sorry for the confusion :facepalm: thanks for clarifying! I'm good with closing the PR.

PR #152, opened 2022-01-07
Fix accidental erasure of finishing request

### LINKED ISSUE #138: finishing task wont get re-added after task reassignment
## Issue

I will use a rmf_demo world to illustrate this issue.

In the new `rmf demos` hotel world, there are 2 cleaning robots operating under a fleet adapter. `Park` task is chosen as the finishing task of the cleaner robots. The expectation here is, cleaning robots will head back to their respective parking spots after finishing their tasks.

run:
```
ros2 launch rmf_demos_gz hotel.launch.xml
```

Then, we will submit 2 cleaning tasks to the robot, when submit the first task
```bash
ros2 run rmf_demos_tasks dispatch_clean -cs clean_lobby --use_sim_time
ros2 run rmf_demos_tasks dispatch_clean -cs clean_waiting_area --use_sim_time
```

---

1. When submit first clean task, assignment:
```bash
 --Agent: 0
 --Agent: 1
     <Clean0: 43676000000.000000, 43725000000.000000, 134054395504.000000, 99.190534%>
     <ParkRobotf0627c: 134054395504.000000, 134054395504.000000, 179361137360.000000, 98.808748%>

#Tasks queue (active | pending)
Robot1: X | Clean0, ParkRobotf0627c
Robot2: X | X
```

Immediately, the first `Clean0` task tuns from pending to active.

2. Submit 2nd clean task.  Since only pending tasks can get reassigned, `Clean0` is not known to the task planner. In this scenario, `Clean1` task is allocated to a different robot. Here we will see that the previously added Park task is removed and not get re-added again.
```bash
assignments:
 --Agent: 0
     <Clean1: 67879000000.000000, 67942000000.000000, 198391058556.000000, 98.233099%>
     <ParkRobot44e59d: 198391058556.000000, 198391058556.000000, 289777678867.000000, 97.389582%>
 --Agent: 1

#Tasks queue (active | pending)
Robot1: Clean0 | 
Robot2: Clean1 | ParkRobot44e59d
```

Expected outcome: The task planner will append a `Park` task to agents that have (user submitted) task(s) in their task queues, regardless of active/pending :
```bash
# Tasks queue (active | pending)
Robot1: Clean0 | ParkRobot90834
Robot2: Clean1 | ParkRobot44e59d
```

Yadunund:
Fix #138

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -142,7 +142,10 @@ void TaskManager::set_queue(
   // function that is called at the end of this function.
   {
     std::lock_guard<std::mutex> guard(_mutex);
-    _queue.clear();
+    // If the queue only has a single automatic task then we do not clear as
+    // this may be a finishing request
+    if (!(_queue.size() == 1 && _queue.back()->request()->automatic()))
+      _queue.clear();
 
     // We use dynamic cast to determine the type of request and then call the
     // appropriate make(~) function to convert the request into a task
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -144,8 +144,12 @@ void TaskManager::set_queue(
     std::lock_guard<std::mutex> guard(_mutex);
     // If the queue only has a single automatic task then we do not clear as
     // this may be a finishing request
-    if (!(_queue.size() == 1 && _queue.back()->request()->automatic()))
+    if (!(assignments.empty() &&
+      _queue.size() == 1 &&
+      _queue.back()->request()->automatic()))
+    {
       _queue.clear();
+    }
 
     // We use dynamic cast to determine the type of request and then call the
     // appropriate make(~) function to convert the request into a task


youliangtan:
Tested it and both robots manage to run the finishing task as intended. While viewing the tasks on the panel, the park task is shown as `canceled` because the dispatcher is also keeping track of the task. Similar logic/or filter should also be implemented on the dispatcher node. But for now, I think this can be neglected since we are revamping the entire task system on `redesign_v2`?

One side observation that I noticed, when the robot is executing a cleaning task (aka docking), it seems interruptable, which shouldn't be the case? 

---
On certain occasions, while running some scenarios in the hotel world, "near miss" situation of 2 robots will interrupt the cleaning robot while it is still moving along the cleaning path. This will end the "dock" process, subsequently show a 'fleetadaptererror' log because of the diverged path that it follows.
```
 [cleanerBotA_fleet_adapter]: Fleet driver [cleanerBotA] reported interruption for [cleanerBotA_2]
[full_control-13] [ERROR] [1641781347.416469257] [cleanerBotA_fleet_adapter]: [RobotUpdateHandle::update_position] The robot [cleanerBotA/cleanerBotA_2] has diverged from its navigation graph
```
Screenshot from 2022-01-10 10-22-34

I guess the above mentioned interruption issue should be a different problem.

Yadunund:
Thanks for the quick review @youliangtan 

> While viewing the tasks on the panel, the park task is shown as canceled because the dispatcher is also keeping track of the task.

Yeah with the legacy system, there seems to be two sources of "what are the tasks being performed", one by the dispatcher and the other by the fleet adapter (internal). With the `redesign_v2` implementation, we hope to introduce a single source of this information.
Having said that, I think in your testing, you should have observed that despite the park task being "cancelled" after reassingment, when the robot does correctly begin the task, the task will appear as "Active". 

> On certain occasions, while running some scenarios in the hotel world, "near miss" situation of 2 robots will interrupt the cleaning robot while it is still moving along the cleaning path. This will end the "dock" process, subsequently show a 'fleetadaptererror' log because of the diverged path that it follows.

This problem should be solved with the new fleet adapter implementations https://github.com/open-rmf/rmf_demos/pull/109. If you still encounter the problem, let's open a separate ticket for that.

youliangtan (approved):
LGTM then. On that separate issue, let me test it out again, will open an issue in `rmf_demos` if it resurfaces again.

PR #155, opened 2022-01-15
Support PerformAction event in composed tasks

Yadunund:
This PR aims to introduce support for `PerformAction` event. Changes
1) JSON schema for `PerformAction` event description
2) `FleetUpdateHandle::add_performable_action(const std::string& category) API: Set the category of actions that robots in this fleet can perform.`
3) `RobotUpdateHandle::set_action_executor(ActionExecutor action_executor)` API: Set the executor which can be used by the fleet adapter to request a robot to perform an action and `RobotUpdateHandle::update_action_remaining_time(rmf_traffic::Duration duration)` to update the estimate of the time remaining for the action.
4) `Standby` and `Active` definitions for `PerformAction`

This PR requires https://github.com/open-rmf/rmf_task/pull/49

### DIFF
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -128,6 +128,7 @@ public:
     // TODO: Consider giving access to the participant schedule and
     // traffic negotiation
 
+    // The desctructor will trigger finished() if it has not already been called
     ~ActionExecution();
 
     class Implementation;
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -128,11 +128,7 @@ public:
     // TODO: Consider giving access to the participant schedule and
     // traffic negotiation
 
-    ~ActionExecution()
-    {
-      // Automatically trigger finished when this object dies
-      finished();
-    }
+    ~ActionExecution();
 
     class Implementation;
   private:
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -122,12 +122,8 @@ public:
     // Trigger this when the action is finished
     void finished();
 
-    // Allow the action performer to report directly to the schedule.
-    // This returns nullptr after `finished()` has been triggered.
-    rmf_traffic::schedule::Participant* schedule();
-    const rmf_traffic::schedule::Participant* schedule() const;
-
-    // TODO: Consider giving access to the traffic negotiation
+    // TODO: Consider giving access to the participant schedule and
+    // traffic negotiation
 
     ~ActionExecution()
     {
@@ -138,10 +134,8 @@ public:
     class Implementation;
   private:
     ActionExecution();
-    rmf_utils::unique_impl_ptr<Implementation> _pimpl;
+    rmf_utils::impl_ptr<Implementation> _pimpl;
   };
-
-  using ActionExecutionPtr = std::shared_ptr<ActionExecution>;
 
   /// Signature for a callback to request the robot to perform an action
   ///
@@ -151,13 +145,13 @@ public:
   /// \param[in] description
   ///   A description of the action to be performed
   ///
-  /// \param[in] completed
+  /// \param[in] execution
   ///   An ActionExecution object that will be provided to the user for
-  ///   managing the state of the action.
+  ///   updating the state of the action.
   using ActionExecutor = std::function<void(
     const std::string& category,
     const nlohmann::json& description,
-    ActionExecutionPtr execution)>;
+    ActionExecution execution)>;
 
   /// Set the ActionExecutor for this robot
   void set_action_executor(ActionExecutor action_executor);
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -110,6 +110,11 @@ public:
   /// value that was given to the setter.
   rmf_utils::optional<rmf_traffic::Duration> maximum_delay() const;
 
+  /// Update the estimate of the time remaining for the action that the robot
+  /// may be performing
+  void update_action_remaining_time(
+  const rmf_traffic::Duration remaining_time_estimate);
+
   class Implementation;
 
   /// This API is experimental and will not be supported in the future. Users
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -122,6 +122,9 @@ public:
     // Trigger this when the action is finished
     void finished();
 
+    // Returns false if the Action has been killed or cancelled
+    bool okay() const;
+
     // TODO: Consider giving access to the participant schedule and
     // traffic negotiation
 
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -117,7 +117,10 @@ public:
 
   /// Signature for a callback to request the robot to perform an action
   ///
-  /// \param[in] action
+  /// \param[in] category
+  ///   A category of the action to be performed
+  ///
+  /// \param[in] description
   ///   A description of the action to be performed
   ///
   /// \param[in] completed
@@ -125,8 +128,9 @@ public:
   ///   robot has completed the action
   using ActionExecutor =
     std::function<void(
-      const nlohmann::json& action,
-      const ActionCompleted& completed)
+      const std::string& category,
+      const nlohmann::json& description,
+      ActionCompleted completed)
     >;
 
   /// Set the ActionExecutor for this robot
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -111,9 +111,37 @@ public:
   /// value that was given to the setter.
   rmf_utils::optional<rmf_traffic::Duration> maximum_delay() const;
 
-  /// Signature for a callback to be executed to let the fleet adapter know that
-  /// the robot has completed the requested action
-  using ActionCompleted = std::function<void()>;
+  /// The ActionExecution class should be used to manage the execution of and
+  /// provide updates on ongoing actions.
+  class ActionExecution
+  {
+  public:
+    // Update the amount of time remaining for this action
+    void update_remaining_time(rmf_traffic::Duration remaining_time_estimate);
+
+    // Trigger this when the action is finished
+    void finished();
+
+    // Allow the action performer to report directly to the schedule.
+    // This returns nullptr after `finished()` has been triggered.
+    rmf_traffic::schedule::Participant* schedule();
+    const rmf_traffic::schedule::Participant* schedule() const;
+
+    // TODO: Consider giving access to the traffic negotiation
+
+    ~ActionExecution()
+    {
+      // Automatically trigger finished when this object dies
+      finished();
+    }
+
+    class Implementation;
+  private:
+    ActionExecution();
+    rmf_utils::unique_impl_ptr<Implementation> _pimpl;
+  };
+
+  using ActionExecutionPtr = std::shared_ptr<ActionExecution>;
 
   /// Signature for a callback to request the robot to perform an action
   ///
@@ -124,22 +152,15 @@ public:
   ///   A description of the action to be performed
   ///
   /// \param[in] completed
-  ///   An ActionCompleted callback that should be called by the user when the
-  ///   robot has completed the action
-  using ActionExecutor =
-    std::function<void(
-      const std::string& category,
-      const nlohmann::json& description,
-      ActionCompleted completed)
-    >;
+  ///   An ActionExecution object that will be provided to the user for
+  ///   managing the state of the action.
+  using ActionExecutor = std::function<void(
+    const std::string& category,
+    const nlohmann::json& description,
+    ActionExecutionPtr execution)>;
 
   /// Set the ActionExecutor for this robot
   void set_action_executor(ActionExecutor action_executor);
-
-  /// Update the estimate of the time remaining for the action that the robot
-  /// may be performing
-  void update_action_remaining_time(
-    const rmf_traffic::Duration remaining_time_estimate);
 
   class Implementation;
 
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -25,6 +25,7 @@
 #include <rmf_traffic/schedule/Participant.hpp>
 
 #include <Eigen/Geometry>
+#include <nlohmann/json.hpp>
 
 #include <vector>
 #include <memory>
@@ -109,6 +110,27 @@ public:
   /// it may take some time for the return value of this getter to match the
   /// value that was given to the setter.
   rmf_utils::optional<rmf_traffic::Duration> maximum_delay() const;
+
+  /// Signature for a callback to be executed to let the fleet adapter know that
+  /// the robot has completed the requested action
+  using ActionCompleted = std::function<void()>;
+
+  /// Signature for a callback to request the robot to perform an action
+  ///
+  /// \param[in] action
+  ///   A description of the action to be performed
+  ///
+  /// \param[in] completed
+  ///   An ActionCompleted callback that should be called by the user when the
+  ///   robot has completed the action
+  using ActionExecutor =
+    std::function<void(
+      const nlohmann::json& action,
+      const ActionCompleted& completed)
+    >;
+
+  /// Set the ActionExecutor for this robot
+  void set_action_executor(ActionExecutor action_executor);
 
   /// Update the estimate of the time remaining for the action that the robot
   /// may be performing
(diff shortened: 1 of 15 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -974,6 +978,76 @@ void FleetUpdateHandle::Implementation::add_standard_tasks()
...
+
+  const auto deserializer =
+    [validator, place = _pimpl->deserialization.place](
+      const nlohmann::json& msg) -> DeserializedEvent
+    {
+      validator->validate(msg);
+      try
+      {
+        const nlohmann::json action =
+          nlohmann::json::parse(msg["action"].get<std::string>());
```
mxgrey, line 1001: I don't think this is a good way to use the `nlohmann` library or JSON more generally. What this is literally saying is that the `action` field will refer to a string, and that string will contain an entire JSON message inside of it. So that might look something like this:

```
"action": "{
  \"clean\": {
    \"zone\": \"terminal_3\"
  }
}"
```

It would become a nightmare of escaping quotes and a very confusing structure.

I recommend the following schema instead:

```
"perform_action": {
  "category": "some_simple_string",
  "description": {
    "any": 0
    "details": { },
    "about": [ ],
    "action: " "
  },
  "expected_finish_location": <optional>,
  "unix_millis_action_duration_estimate": <optional>,
  ... any other optional task planner details ...
}
```

The `add_performable_action` function will have this signature instead:

```
FleetUpdateHandle& FleetUpdateHandle::add_performable_action(
  const std::string& category,
  ConsiderRequest consider);
```

We will maintain a map from `category` keys to `consider` callbacks. When a `perform_action` event comes in, we will do a look up on that map. If the lookup fails, we reject the task. If the look up succeeds, we trigger the `consider` callback and then accept or reject based on its response.

The `ActionExecutor` signature would look like this instead:

```
 using ActionExecutor =
    std::function<void(
      const std::string& category,
      const nlohmann::json& description,
      ActionCompleted completed)
```

This should make it very easy for users to put in their own arbitrary JSON details for the action while still allowing the fleet adapter to do the heavy lifting in weeding out irrelevant task requests.

mxgrey (changes requested):

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Thanks for the feedback. I can see how this is much easier for the user. I have a couple of questions on the schema
1) What is the purpose of the `any`, `detail` and `about` fields?
2) I assume the intention is for the entire `description` JSON block to be passed into `PerformAction::Description::make()` as the `action`. Is this correct?

mxgrey: The `any`, `details`, `about`, `this`, `action` are just examples of how a user can fill the `description` JSON with any kind of data they want. You can totally disregard them, they're not actually part of the schema, just placeholders.

> I assume the intention is for the entire description JSON block to be passed into PerformAction::Description::make() as the action. Is this correct?

That's right, the `PerformAction::Description` would need to have some kind of `action_message` field (or something similarly named) to store the message that gets passed to the robot command handle.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -110,6 +111,36 @@ class RobotUpdateHandle
...
+      ActionCompleted completed)
+    >;
+
+  /// Set the ActionExecutor for this robot
+  void set_action_executor(ActionExecutor action_executor);
+
+  /// Update the estimate of the time remaining for the action that the robot
+  /// may be performing
+  void update_action_remaining_time(
+  const rmf_traffic::Duration remaining_time_estimate);
```
mxgrey, line 142: Nitpick on spacing. Also `const` doesn't do anything here.

rmf_fleet_adapter/schemas/event_description_PerformAction.json
```diff
@@ -0,0 +1,30 @@
...
+    "category": {
+      "description": "A string that describes the nature of this action. It should match that passed into FleetUpdateHandle::add_performable_action()",
+      "type": "string"
+    },
+    "description": {
+      "description": "A dictionary describing the action to be performed.",
+      "type": "object"
+    },
+    "unix_millis_action_duration_estimate": {
+      "description": "(Optional) The estimated duration for this action",
```
mxgrey, line 17: I wonder if we should change `(Optional)` to `(Recommended)`. If the planner assumes 0 time for a looong action, then the planning and dispatching could get really messed up.

mxgrey: Alternatively, we could have some way for system integrators to give the fleet adapter an estimate of the time (maybe also the finish location) based on the information provided in the description.

But these are really thoughts for future enhancements, not really important right now.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -110,6 +111,36 @@ class RobotUpdateHandle
   /// value that was given to the setter.
   rmf_utils::optional<rmf_traffic::Duration> maximum_delay() const;
 
+  /// Signature for a callback to be executed to let the fleet adapter know that
+  /// the robot has completed the requested action
+  using ActionCompleted = std::function<void()>;
```
mxgrey, line 116: I see that things like `ActionCompleted` and `update_action_remaining_time` are separated right now even though they're conceptually coupled together and both coupled with the execution of a specific action instance. Instead of having them decoupled in the RobotUpdateHandle interface, how about this:

```
class ActionExecution
{
public:
  // Update the amount of time remaining for this action
  void update_remaining_time(rmf_traffic::Duration remaining_time_estimate);

  // Trigger this when the action is finished
  void finished();

  // Allow the action performer to report directly to the schedule.
  // This returns nullptr after `finished()` has been triggered.
  Participant* schedule();
  const Participant* schedule() const;

  // TODO: Consider giving access to the negotiation? Maybe in a future version

  ~ActionExecution()
  {
    // Automatically trigger finished when this object is lost
    finished();
  }

  class Implementation;
private:
  rmf_utils::unique_impl_ptr<Implementation> _pimpl;
};

using ActionExecutor = std::function<void(
  const std::string& category,
  const nlohmann::json& description,
  ActionExecution execution)>;
```

This pattern will allow us to extend the action execution API in the future (e.g. giving action executors the ability to negotiate traffic) while remaining backwards compatible.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
```diff
@@ -0,0 +1,248 @@
...
+rmf_traffic::Duration PerformAction::Active::remaining_time_estimate() const
+{
+  if (_context->action_remaining_time().has_value())
+  {
+    return _context->action_remaining_time().value();
+  }
+
+  // If an estimate is not provided we compute one based on the expected finish
+  // time
+  return _expected_finish_time - _context->now();
```
mxgrey, line 176: We should be careful about this going negative. Maybe do `std::max(0, _expected_finish_time - _context->now());`

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+//==============================================================================
+auto PerformAction::Active::backup() const -> Backup
+{
+  // PerformAction doesn't need to be backed up
```
mxgrey, line 182: In the future we could consider adding a function to `ActionExecution` that allows the system integrator to periodically send in backup messages. But for now, I think this is fine.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
```diff
@@ -0,0 +1,248 @@
...
+  _state->update_dependencies({});
+
+  _context->worker().schedule(
+    [task_is_interrupted](const auto&)
+    {
+      task_is_interrupted();
+    });
+
+  // TODO(YV): Currently we will ask the robot to perform the action again when
+  // resumed. Future work can be to receive a callback from RobotUpdateHandle
```
mxgrey, line 201: Let's receive that callback in the `ActionExecution` class that I've proposed. That can be future work, though.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
```diff
@@ -0,0 +1,248 @@
...
+      }
+    });
+}
+
+//==============================================================================
+void PerformAction::Active::cancel()
+{
+  _state->update_status(Status::Canceled);
+  _state->update_log().info("Received signal to cancel");
+  _finished();
```
mxgrey, line 218: We should think about ways to signal to the `ActionExecution` that it has been cancelled. Maybe give it a `bool okay() const` function, and set the `schedule` to `nullptr`? Maybe have it accept a callback that we can trigger when a cancellation happens?

mxgrey (changes requested):
This is generally looking really good! I just have some recommendations for tightening up the API.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
Yadunund: Done https://github.com/open-rmf/rmf_ros2/pull/155/commits/9a8ad648fa3b2baa07c273ea4ccfc5d941576c29

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
Yadunund: I've updated the comment to leave this for future work.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
Yadunund: Noted. Updated the comment.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
Yadunund: Done

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
Yadunund: Implemented as discussed.

rmf_fleet_adapter/schemas/event_description_PerformAction.json
Yadunund: I've updated to `Recommended`.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -974,6 +981,95 @@ void FleetUpdateHandle::Implementation::add_standard_tasks()
...
+        std::vector<std::string> errors = {};
+        return {description, errors};
+      }
+      catch(const std::exception& e)
+      {
+        return {nullptr, {e.what()}};
+      }
+    };
+
+  _pimpl->deserialization.event->add(
```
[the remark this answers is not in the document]
Yadunund: Moved it into `FleetUpdateHandle::Implementation::make(~)`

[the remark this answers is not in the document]
Yadunund: Moved it into `FleetUpdateHandle::Implementation::make(~)`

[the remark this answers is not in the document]
mxgrey: Sorry to be nitpicky about this, but could we imitate the pattern that's being used in [`FleetUpdateHandle::Implementation::add_standard_tasks`](https://github.com/open-rmf/rmf_ros2/blob/redesign_v2/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L932)? Basically put this implementation into a `rmf_tasks::add_perform_action(~)` function and define that function in src/rmf_fleet_adapter/tasks/PerformAction.cpp`?

I think that would organize the code a bit better and reduce the already considerable amount of noise in this `make` function.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -110,6 +111,54 @@ class RobotUpdateHandle
...
+    // Trigger this when the action is finished
+    void finished();
+
+    // Returns false if the Action has been killed or cancelled
+    bool okay() const;
+
+    // TODO: Consider giving access to the participant schedule and
+    // traffic negotiation
+
+    ~ActionExecution()
```
mxgrey, line 131: We should put the definition of this function into the proper translation unit, i.e. `RobotUpdateHandle.cpp`. I understand you were probably copying the example code that I provided, but I defined it inline for brevity. It's not actually good to leave this function definition exposed in a public API, because that can lead to poorly defined behavior if we ever change the implementation of the function.

rmf_fleet_adapter/schemas/event_description_PerformAction.json
```diff
@@ -0,0 +1,30 @@
...
+  "description": "Description schema for the perform_action category",
+  "type": "object",
+  "properties": {
+    "category": {
+      "description": "A string that describes the nature of this action. It should match that passed into FleetUpdateHandle::add_performable_action()",
+      "type": "string"
+    },
+    "description": {
+      "description": "A dictionary describing the action to be performed.",
+      "type": "object"
```
mxgrey, line 14: Let's not force this to be a dictionary. Maybe a system integrator would like this to be an array. We can just remove this `"type": "object"` specification and leave it ambiguous.

I'd also recommend changing `description` to something like `"Additional information that will be passed along to the action executor."

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
```diff
@@ -393,6 +393,20 @@ uint32_t RobotContext::current_mode() const
   return _current_mode;
 }
 
+//==============================================================================
+void RobotContext::action_executor(
+  RobotUpdateHandle::ActionExecutor action_executor)
+{
+  if (action_executor != nullptr)
```
mxgrey, line 400: I'm not actually sure that we should block this. What if the user is trying to erase its action executor for some reason?

I recommend instead of blocking, we simply emit a warning that the action executor is being nullified which will lead to an automatic critical task error if a task requires an action to be performed while this is still set to null.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_RobotUpdateHandle.hpp
```diff
@@ -40,26 +74,25 @@ class RobotUpdateHandle::Implementation
...
   }
 
-  static Implementation& get(RobotUpdateHandle& handle)
+  static Implementation &get(RobotUpdateHandle &handle)
   {
     return *handle._pimpl;
   }
 
-  static const Implementation& get(const RobotUpdateHandle& handle)
+  static const Implementation &get(const RobotUpdateHandle &handle)
```
mxgrey, line 88: Looks like some automatic styling has messed things up here.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
```diff
@@ -0,0 +1,269 @@
...
+void PerformAction::Active::_execute_action()
+{
+  auto action_executor = _context->action_executor();
+  if (action_executor == nullptr)
+  {
+    // The action_executor has not been set by the user
+    _state->update_status(Status::Error);
+    const std::string msg = "ActionExecutor not set via RobotUpdateHandle. "
+      "Unable to perform the requested action.";
+    _state->update_log().info(msg);
```
mxgrey, line 251: This should be an error-level log, not info.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/PerformAction.cpp
```diff
@@ -0,0 +1,269 @@
...
+    _finished();
```
mxgrey, line 252: Let's update the status to an error before finishing.

mxgrey:
This is shaping up very nicely. I just left a few more small remarks to address and then I'll be happy to approve :ok_hand:

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -128,11 +128,7 @@ class RobotUpdateHandle
     // TODO: Consider giving access to the participant schedule and
     // traffic negotiation
 
-    ~ActionExecution()
-    {
-      // Automatically trigger finished when this object dies
-      finished();
-    }
+    ~ActionExecution();
```
mxgrey, line 131: Just one last minor thing: Let's add a line of documentation to this destructor explaining that `finished()` will automatically be triggered if it wasn't triggered already.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
```diff
@@ -272,5 +286,38 @@ void RobotUpdateHandle::Unstable::set_lift_entry_watchdog(
...
+void RobotUpdateHandle::ActionExecution::update_remaining_time(
+    rmf_traffic::Duration remaining_time_estimate)
+{
+  _pimpl->data->remaining_time = remaining_time_estimate;
+}
+
+//==============================================================================
+void RobotUpdateHandle::ActionExecution::finished()
+{
+  _pimpl->data->finished();
```
mxgrey, line 299: Oh, and one other final detail. We should make sure that the finished callback can only get triggered exactly once. We can do that like so:

```
if (!_pimpl->data->finished)
  return;

_pimpl->data->finished();
_pimpl->data->finished = nullptr;
```

Yadunund: Done 👍🏼 
https://github.com/open-rmf/rmf_ros2/pull/155/commits/22a4fbae3018a9f3eaafed614731857c822d5a25

mxgrey (approved):
Looks great! :tada:

PR #158, opened 2022-01-25
Feature: Direct Task Assignment

Yadunund:
This PR introduces the ability to directly assign a task to a robot in a fleet, thereby bypassing the bidding system. Users can publish an `rmf_task_msgs::msg::ApiRequest`  msg with `json_msg` field following the schema introduced in https://github.com/open-rmf/rmf_api_msgs/pull/15 over the topic `/task_api_requests`

The implementation adds a `_direct_queue` in `TaskManager` which maintains the assignments for direct tasks. One point about this appraoch:
* Maintaining `_queue` and `_direct_queue` introduces the complexity of checking which queue to process in several functions including `_begin_next_task()`, `expected_finish_state()`, `remove_task_from_queue()`. I think things could be a lot more elegant if we just maintained a single `_queue`. We could do this if the Task Planner was capable of generating assignments where direct tasks are assigned before dispatched ones. One way of doing this would be to introduce a new priority scheme which also incorporates the existing binary scheme. Direct tasks would be automatically assigned a "high" priority such that they are assigned before dispatched ones. And the binary high/low priority could till apply within the dispatched and assigned tasks. Another advantage of this would be that the planner's time segmentation would allow for a natural distribution of the dispatch and direct tasks. So there wouldn't be a confusion on whether to begin a dispatch or direct task next.

But right now, I'm trying to make things work with the separate queues and have the following questions:
    1) Initially I thought the direct assignments should be in order that they were received. But if that is the case, it could be troublesome if a user submitted a direct task to begin one hour from now first and then another direct task to begin immediately. But given this sequence, the first task will be at the top of the queue and the second one will only begin after this one completes- which is unproductive. Hence, i'm currently replanning with all the task in `_direct_queue` as this would naturally time segment the assignments. 
    2) Decision on which task to begin: Say we have tasks queued in both `_queue` and `_direct_queue`. The top of the `direct_queue` has a deployment time in the future while the one in `_queue` is ready to be deployed. Should we start the one in `_queue` or always complete all the `_direct_queue` tasks before commencing the dispatch ones? This could happen in a scenario where the fleet has already been awarded a dispatch task (which is now queued) and then receives a direct task request after. 
    
A lot of these problems can be solved trivially if we have a configuration option for the fleet adapter to respond to either `dispatch` or `direct` task requests but not both. Mixing the two capabilities is tricky as the behaviour may favour one deployment scenario but not the other.

Please let me know your thoughts @mxgrey  @cnboonhan

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -821,6 +821,13 @@ void TaskManager::set_queue(
   // function that is called at the end of this function.
   {
     std::lock_guard<std::mutex> guard(_mutex);
+    // Do not remove automatic task if assignments is empty. See Issue #138
+    if (assignments.empty() &&
+      _queue.size() == 1 &&
+      _queue.front().request()->booking()->automatic())
+    {
+      return;
+    }
     _queue = assignments;
     _publish_task_queue();
   }
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -780,15 +780,7 @@ auto TaskManager::expected_finish_state() const -> State
 
   if (!_direct_queue.empty())
   {
-    rmf_task::State finish_state;
-    auto queue = _direct_queue;
-    while (!queue.empty())
-    {
-      const auto top = queue.top();
-      finish_state = top.assignment.finish_state();
-      queue.pop();
-    }
-    return finish_state;
+    return _direct_queue.rbegin()->assignment.finish_state();
   }
 
   if (_active_task)
@@ -913,19 +905,18 @@ void TaskManager::_begin_next_task()
     return;
   }
 
+  // The next task should one in the direct assignment queue if present
+  const bool is_next_task_direct = _direct_queue.empty() ? false : true;
+  const auto assignment = is_next_task_direct ?
+    _direct_queue.begin()->assignment :
+    _queue.front();
+
   // We take the minimum of the two to deal with cases where the deployment_time
   // as computed by the task planner is greater than the earliest_start_time
   // which is greater than now. This can happen for example if the previous task
   // completed earlier than estimated.
   // TODO: Reactively replan task assignments across agents in a fleet every
   // time as task is completed.
-
-  // The next task should one in the direct assignment queue if present
-
-  const bool is_next_task_direct = _direct_queue.empty() ? false : true;
-  const auto assignment = is_next_task_direct ? _direct_queue.top().assignment
-    : _queue.front();
-
   const auto deployment_time = std::min(
     assignment.request()->booking()->earliest_start_time(),
     assignment.deployment_time());
@@ -949,7 +940,7 @@ void TaskManager::_begin_next_task()
       _task_finished(id));
 
     if (is_next_task_direct)
-       _direct_queue.pop();
+       _direct_queue.erase(_direct_queue.begin());
    else
     _queue.erase(_queue.begin());
 
@@ -1484,9 +1475,22 @@ void TaskManager::_send_simple_error_if_queued(
   const std::string& request_id,
   const std::string& type)
 {
+  // TODO(YV): We could cache the task_ids of direct and dispatched tasks in
+  // unordered_sets and perform a lookup to see which queue to iterate.
   for (const auto& a : _queue)
   {
     if (a.request()->booking()->id() == task_id)
+    {
+      return _send_simple_error_response(
+        request_id, 6, "Invalid Circumstances",
+        type + " a task that is queued (not yet active) "
+        "is not currently supported");
+    }
+  }
+
+  for (const auto& a : _direct_queue)
+  {
+    if (a.assignment.request()->booking()->id() == task_id)
     {
       return _send_simple_error_response(
         request_id, 6, "Invalid Circumstances",
@@ -1650,7 +1654,7 @@ std::vector<std::string> get_labels(const nlohmann::json& request)
 }
 
 //==============================================================================
-void remove_task_from_queue(
+bool remove_task_from_queue(
   const std::string& task_id,
   std::vector<TaskManager::Assignment>& queue)
 {
@@ -1666,9 +1670,33 @@ void remove_task_from_queue(
     if (it->request()->booking()->id() == task_id)
     {
       queue.erase(it);
-      break;
+      return true;
     }
   }
+  return false;
+}
+
+//==============================================================================
+bool remove_task_from_queue(
+  const std::string& task_id,
+  TaskManager::DirectQueue& queue)
+{
+  // If the task is queued, then we should make sure to remove it from the
+  // queue, just in case it reaches an active state before the dispatcher
+  // issues its cancellation request.
+  //
+  // TODO(MXG): We should do a much better of job of coordinating these
+  // different moving parts in the system. E.g. who is ultimately responsible
+  // for issuing the response to the request or updating the task state?
+  for (auto it = queue.begin(); it != queue.end(); ++it)
+  {
+    if (it->assignment.request()->booking()->id() == task_id)
+    {
+      queue.erase(it);
+      return true;
+    }
+  }
+  return false;
 }
 } // namespace anonymous
 
@@ -1748,7 +1776,7 @@ void TaskManager::_handle_direct_request(
         deployment_time)
     };
   ++_next_sequence_number;
-  _direct_queue.push(assignment);
+  _direct_queue.insert(assignment);
 }
 
 //==============================================================================
@@ -1770,8 +1798,10 @@ void TaskManager::_handle_cancel_request(
     _task_state_update_available = true;
     return _send_simple_success_response(request_id);
   }
-
-  remove_task_from_queue(task_id, _queue);
+  // TODO(YV): We could cache the task_ids of direct and dispatched tasks in
+  // unordered_sets and perform a lookup to see which function to call.
+  if (!remove_task_from_queue(task_id, _queue))
+    remove_task_from_queue(task_id, _direct_queue);
 }
 
 //==============================================================================
@@ -1794,7 +1824,10 @@ void TaskManager::_handle_kill_request(
     return _send_simple_success_response(request_id);
   }
 
-  remove_task_from_queue(task_id, _queue);
+  // TODO(YV): We could cache the task_ids of direct and dispatched tasks in
+  // unordered_sets and perform a lookup to see which function to call.
+  if (!remove_task_from_queue(task_id, _queue))
+    remove_task_from_queue(task_id, _direct_queue);
 }
 
 //==============================================================================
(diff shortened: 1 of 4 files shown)


mxgrey:
> Initially I thought the direct assignments should be in order that they were received. But if that is the case, it could be troublesome if a user submitted a direct task to begin one hour from now first and then another direct task to begin immediately. 

Yes, this is a vulnerability of the FIFO approach which we discussed with our collaborators. The conclusion from that discussion was that it's the requester's responsibility to make sure that tasks are submitted in the correct order. I think the intended use case was that the requester would just stream in the requests right when the task should be performed, and the time field would be virtually irrelevant.

Admittedly that does not jive well with the ability for the requester to also set a start time. With that in mind, we could do something slightly different than FIFO: Whenever the robot is available, we choose the task with the lowest start time. Ties will be broken according to the FIFO order that the requests were received in.

I don't believe we should be using the task planner, since the purpose of this direct request pipeline was to skip all optimization, as we expect the requester to have already optimized. I don't think we should go with a priority scheme, since even just reordering the direct requests according to our own optimization criteria could interfere with the behavior that the requester intended.

> Decision on which task to begin: Say we have tasks queued in both _queue and _direct_queue. The top of the direct_queue has a deployment time in the future while the one in _queue is ready to be deployed. Should we start the one in _queue or always complete all the _direct_queue tasks before commencing the dispatch ones? 

> A lot of these problems can be solved trivially if we have a configuration option for the fleet adapter to respond to either dispatch or direct task requests but not both.

Based on our discussions with collaborators, the behavior that's wanted for now is to simply not perform any dispatched requests on a robot until all the robot's direct requests are finished. This does pose a risk that dispatched requests could be needlessly stuck or delayed, but for now we are viewing that as an inherent risk of trying to mix these systems together. I think we can consider mitigation strategies (e.g. send a task request back to the dispatcher to be reconsidered) in a later release after this first draft of the feature.

Yadunund:
Thanks for the feedback @mxgrey .

> I don't believe we should be using the task planner, since the purpose of this direct request pipeline was to skip all optimization, as we expect the requester to have already optimized. I don't think we should go with a priority scheme, since even just reordering the direct requests according to our own optimization criteria could interfere with the behavior that the requester intended.

I assumed we would still want to auto-include `ChargeBattery` tasks hence, I'm using the TaskPlanner to plan with 1 agent (the asigned robot) and the requested task. Do I assume `ChargeBattery` tasks will also be sent over via the direct task pipeline and hence skip using the task planner? 


> With that in mind, we could do something slightly different than FIFO: Whenever the robot is available, we choose the task with the lowest requested start time. Ties will be broken according to the FIFO order that the requests were received in.


> Based on our discussions with collaborators, the behavior that's wanted for now is to simply not perform any dispatched requests on a robot until all the robot's direct requests are finished.

Noted. I'll update the implementation to follow this behavior.

mxgrey:
> I assumed we would still want to auto-include ChargeBattery tasks hence, I'm using the TaskPlanner to plan with 1 agent (the asigned robot) and the requested task. Do I assume ChargeBattery tasks will also be sent over via the direct task pipeline and hence skip using the task planner?

This is a tricky question. IIRC the task manager has a behavior where if the robot is idle and the battery is below a threshold then it will automatically tell the robot to charge, correct? If so, then we can start out by relying on this combined with responsible separation of tasks requests by the requester to keep the battery charged.

Admittedly that's not a robust system, so in a future release we could make some changes to the task planner that allow us to constrain the order that tasks occur in. If we tell the task planner to constrain the order of all direct tasks then the planner's output will simply be those tasks in their original order but with charge battery tasks mixed in. But I would save this for future work.

Yadunund:
> This is a tricky question. IIRC the task manager has a behavior where if the robot is idle and the battery is below a threshold then it will automatically tell the robot to charge, correct? If so, then we can start out by relying on this combined with responsible separation of tasks requests by the requester to keep the battery charged.

Yes should be possible. Just need to tweak the implementation to add this task to the direct task queue.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -163,84 +162,23 @@ std::string make_error_str(
...
   {
-    return respond(
-      {
-        std::nullopt,
-        {make_error_str(
-          4, "Unsupported type",
-          "Fleet [" + name + "] does not support task category ["
-          + category + "]")}
-      });
+    errors->push_back(make_error_str(
```
mxgrey, line 177: Are we guaranteed to always be provided a non-null `errors`? If so, why not pass in `std::vector<std::string>&` instead? If not, we need to check `if (errors)` here and also in all the places that we use `errors` below.

mxgrey (changes requested):
I've left a few remarks to look over.

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -771,22 +771,35 @@ std::string TaskManager::robot_status() const
...
 
-  // Update battery soc and finish time in the current state
-  auto finish_state = _context->current_task_end_state();
-  auto location = finish_state.extract_plan_start().value();
-  finish_state.time(rmf_traffic_ros2::convert(_context->node()->now()));
+  if (!for_direct_assignment)
+  {
+    if (!_active_task)
+    {
+      // There may be queued up direct tasks but since none of them are active
```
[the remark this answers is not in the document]
Yadunund: I haven't pushed my latest changes yet but this code from my earlier understanding. The idea was if there are no active tasks, then perhaps the direct task is scheduled to begin much later and hence queued. So return the current state as the robot may be able to squeeze this dispatch task before the direct ask needs to be deployed.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: I should be checking `if (errors)` which I missed out. Not using `std::vector<std::string>&` to retain the `const` correctness of the function

mxgrey: I don't see how the choice between a pointer and a reference would impact const correctness. The only effective difference between the two is that references are not (supposed to be) nullable. If we always expect the caller of the function to provide a valid reference to a `std::vector<std::string>` object, then we should give the function a mutable reference argument `std::vector<std::string>&`.

Yadunund:
@mxgrey this PR is ready for a review.

Test results:
If we submit a `patrol` task via the dispatcher, it gets assigned to `tinyRobot1` 
```
ros2 run rmf_demos_tasks dispatch_loop -s coe -f pantry -n 10 --use_sim_time
```
https://user-images.githubusercontent.com/13482049/151151956-d06f0c77-1a55-4941-bbcc-2e72c1bedd10.mp4

But if we submit the task via a direct api request, it gets assigned to the robot requested, `tinyRobot2`. The script I'm using can be found in `rmf_demos/redesign_v2`.
```
ros2 run rmf_demos_tasks direct_patrol -F tinyRobot -R tinyRobot2 -s coe -f pantry -n 10 --use_sim_time
```

https://user-images.githubusercontent.com/13482049/151152922-c6494ed9-c9e5-4c93-b647-66069ea56e12.mp4

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Whoops I was totally wrong here. Must have gotten confused with an unrelated error.

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -898,26 +917,31 @@ void TaskManager::_begin_next_task()
     return;
   }
 
-  const rmf_traffic::Time now = rmf_traffic_ros2::convert(
-    _context->node()->now());
-  const auto next_task = _queue.front();
+  // The next task should one in the direct assignment queue if present
+  const bool is_next_task_direct = _direct_queue.empty() ? false : true;
```
mxgrey, line 921: Nitpick: I think it would be simpler to just say

```
const bool is_next_task_direct = !_direct_queue.empty();
```

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -927,6 +951,9 @@ void TaskManager::_begin_next_task()
       _phase_finished_cb(),
       _task_finished(id));
 
+    if (is_next_task_direct)
+       _direct_queue.erase(_direct_queue.begin());
+   else
```
mxgrey, line 956: Nitpick: There's some inconsistent indentation through here.

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.hpp
```diff
@@ -190,13 +231,19 @@ class TaskManager : public std::enable_shared_from_this<TaskManager>
   friend class ActiveTask;
 
   agv::RobotContextPtr _context;
-  std::weak_ptr<BroadcastClient> _broadcast_client;
+  std::optional<std::weak_ptr<BroadcastClient>> _broadcast_client;
```
mxgrey, line 234: I'm curious, what's the motive behind having `std::optional<std::weak_ptr<T>>` here? An ordinary `std::weak_ptr` is nullable, so what extra layer of expression do we achieve by wrapping a `std::optional` around it? I don't think there's anything wrong with it; I'm just not sure what it's meant to express.

mxgrey (approved):
I left some minor remarks, but I don't see any of them as blockers.

I really like how the abstract priority system has allowed you to easily get the FIFO outcome out of the task planner without any changes to the task planner itself :+1:

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.hpp
Yadunund: I was trying to come up with a way to distinguish the situation where we we get a `nullptr` when we try to lock the `weak_ptr` vs we're able to lock, but the `BroadcastClient` was never initialized (server_uri not provided) and hence is a `nullptr`. If there is a more elegant way to do this, please let me know.

mxgrey: Ah that makes perfect sense, thanks for the explanation :+1:

mxgrey: Oh, some more thoughts, though... Why would it be a `std::weak_ptr` to the broadcast client? Generally `std::weak_ptr` is used to avoid circular dependencies or in cases where something just wants to observe the lifetime of a resource without interfering with its lifecycle. I don't think either of those is the case here. So maybe this should just generally be a `std::shared_ptr<BroadcastClient>`?

If we have cases where there might be a delay in the availability of a `BroadcastClient` instance, then we can consider using `std::shared_ptr<std::shared_ptr<BroadcastClient>>` so that the inner `std::shared_ptr` can be updated at any time, and everything holding the outer `std::shared_ptr` will automatically get access to the new resource.

These aren't things we need to worry about for this release, but maybe thoughts to consider for the next iteration.

Yadunund: Whoops sorry I merged it too quickly.

Ah yeah the rationale was to deal with the delay in availability of the `BraodcastClient`. I will add a comment based on what you mentioned as a TODO for future reference 👍🏼

PR #159, opened 2022-01-27
Fix Compose Tasks

Yadunund:
Add phase descriptions to builder in task deserializer.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Compose.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Compose.cpp
@@ -126,26 +126,31 @@ void add_compose(
           errors.end(),
           phase_activity.errors.begin(), phase_activity.errors.end());
 
-        std::vector<rmf_task_sequence::Phase::ConstDescriptionPtr> cancel;
-        const auto& on_cancel_json = phase_json["on_cancel"];
-        if (on_cancel_json.is_array())
+        std::vector<rmf_task_sequence::Phase::ConstDescriptionPtr> cancel = {};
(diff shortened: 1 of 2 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Compose.cpp
```diff
@@ -126,10 +126,11 @@ void add_compose(
           errors.end(),
           phase_activity.errors.begin(), phase_activity.errors.end());
 
-        std::vector<rmf_task_sequence::Phase::ConstDescriptionPtr> cancel;
-        const auto& on_cancel_json = phase_json["on_cancel"];
-        if (on_cancel_json.is_array())
+        std::vector<rmf_task_sequence::Phase::ConstDescriptionPtr> cancel = {};
+        const auto cancel_it = phase_json.find("on_cancel");
+        if (cancel_it != phase_json.end() && cancel_it.value().is_array())
```
mxgrey, line 131: We should not silently ignore `on_cancel` if it doesn't map to an array.

We can skip adding the cancellation sequence, but let's at least add a message to the `errors` vector so that the operators can be informed of the issue.

mxgrey (changes requested):
Great fix, but I do have one quick request.

mxgrey (approved):
Oh wait, I just remembered that the schema of `on_cancel` requires it to be an array. If the validator approved of the message, then it should certainly be an array. Nothing needs to be changed here.

mxgrey:
...Although it might not hurt to issue some kind of error anyway, just in case the schema has been changed while the assumptions of this code fail to be updated.

Issue #163, opened 2022-02-03
Avoid robot backtracking when a RobotCommandHandle::follow_new_path() is triggered


Yadunund:
`rmf_fleet_adapter` uses `rxcpp` for reactively generating a new plan for the robot on a number of occasions including during traffic negotiation and when itinerary delays accumulate past a threshold. As a response, the `RobotCommandHandle::follow_new_path()` may be triggered to instruct the robot to follow the updated itinerary. However, between the time of the interruption and when the implementation of `follow_new_path()` sends an instruction to the robot, the robot would have continued moving along its previous path. As a result when the robot follows the new set of waypoints, it "backtracks", ie moves to the first position in the path which is slightly behind the robots current position, before continuing along the remaining path. Presently the workaround for this is to perform some filtering in the waypoints received.

Using this ticket to discuss potential solutions for this problem.

One easy solution would be to request the robot to stop immediately whenever the interrupt is triggered. The `RobotUpdateHandle` could accept a callback from users that would get their robots to stop when triggered. However, this is a big change in the behavior as I think one of the goals with the current implementation is to have the robots continue along their plans while the negotiations occur.
Another possible approach is to have the planner perform some kind of trajectory tuning/optimization based on the kinematics of the fleet.

Looking for feedback on these ideas along with other suggestions.

mxgrey:
I think post-processing of the plan based on the latest robot state is the best approach. We would compare the plan's expectations to our knowledge of the latest physical reality of the robot and just prune some waypoints off the front of the trajectory if the robot has already passed them.

One difficulty with this is identifying cases where the robot *does* **need** to backtrack, e.g. to escape a conflict with another robot. In those cases, pruning the start of the plan would lead to a deadlock. This issue has been the primary blocker that's prevented us from taking action on this problem.

I think when we have the event-driven traffic system that can recognize the purpose behind each wait, we can leverage the event expectation system to identify when the robot **needs** to backtrack versus when it's safe to prune the front of the trajectory.

Yadunund:
Thanks for the valuable insights @mxgrey. Could we look at the `time` property of the first waypoint in the new path to determinate whether the intention is to wait at the previous location? Maybe check if the if the wait duration is above a certain threshold? I assume if the robot is not meant to wait, the first waypoint in the new path will have time very close to `now`?

mxgrey:
[...]

PR #167, opened 2022-02-08
Trigger stop callback during interrupt

xiyuoh:
This PR aims to avoid backtracking when a new path is provided for robots after replanning relevant to #163. THe RobotUpdateHandle accepts a stop callback and stores it in RobotContext. The function gets called when inside `RobotContext::trigger_interrupt()`, stopping the robot so that its current position is more accurate for replanning waypoints. 

Currently only used to store the stop callback in demo fleet adapter.

### DIFF
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -47,6 +47,11 @@ public:
   /// latest position of the robot before calling this function.
   void interrupted();
 
+  /// Store the stop callback in RobotContext to command the robot to stop when
+  /// interrupt is triggered
+  void set_stop_callback(
+    std::function<void(std::string)> stop);
+
   /// Update the current position of the robot by specifying the waypoint that
   /// the robot is on and its orientation.
   void update_position(
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
@@ -180,6 +180,7 @@ RobotContext::observe_interrupt() const
 //==============================================================================
 void RobotContext::trigger_interrupt()
 {
+  this->_stop(this->name());
   _interrupt_publisher.get_subscriber().on_next(Empty{});
 }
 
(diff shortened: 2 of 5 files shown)


mxgrey (changes requested):
The concept of a stop callback sounds really good! There are just two things that I think could be tweaked:
1. Since the callback is being given to each `RobotUpdateHandle`, there's no need for us to pass a robot name to the callback. If the user that's providing the callback needs to know the robot name, they can bind that information into a lambda that they pass in as the callback. In other words, we can just use `std::function<void()>` as the callback signature.
2. We should always check whether the callback is `nullptr` before attempting to trigger it. If it's a `nullptr` then we'll just skip triggering it. This preserves backwards compatibility and also allows users to decide whether or not they want to supply a stop callback.

xiyuoh:
Thank you for the insightful comments! I've made the changes accordingly.

mxgrey (approved):
Thanks, this looks great!

mxgrey (changes requested):
Okay, I'm very sorry to do this, but I just realized that the `RobotCommandHandle` already has a virtual function for [`stop()`](https://github.com/open-rmf/rmf_ros2/blob/14ba7bbfc6da10bcde1e00ea5e4371332525fccc/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotCommandHandle.hpp#L76-L77) that's meant to do exactly what the stop callback should do.

So we should be able to reduce this PR to just

```
void RobotContext::trigger_interrupt()
{
  if (const auto c = command())
    c->stop();
  _interrupt_publisher.get_subscriber().on_next(Empty{});
}
```

We should avoid having redundant ways to command a stop.

xiyuoh:
Hmm so the idea to set a stop callback instead of using the virtual `stop()` function directly is to provide the option of toggling the callback on/off. 

Specifically for the demo fleet adapter, we were experimenting between different stop commands on the fleet manager's end, i.e. stopping very briefly or stopping for 1s. If the stop time is too long, it affects the robots' negotiations and they come to a standstill. I'm thinking it could be beneficial to make it optional, since stop time probably varies for different users.

Do you think we could store a parameter inside `RobotContext` instead that triggers `stop()` only when True?

mxgrey:
Any custom logic related to stopping can be programmed into the implementation of the `RobotCommandHandle::stop()` command.

The same user that could set the stop callback of this PR would also have access to fully customize the virtual `stop()` function of the `RobotCommandHandle`.

For example, you could implement `RobotCommandHandle::stop` so that it checks for a callback that the `RobotCommandHandle` stores for itself and can change at any time.

Let me know if that makes sense. If not, maybe you could point me at a concrete example where you want to use this feature.

xiyuoh:
Ah yes it makes much more sense to check for a callback in `RobotCommandHandle` instead, thank you for pointing it out! Updated the changes in the latest commit

mxgrey (approved):

xiyuoh:
Hello @mxgrey, I had to force push the same changes to sign an old merge commit, so sorry to trouble you to approve again :bow:

mxgrey (approved):

PR #180, opened 2022-02-28
remove unnecessary copy of desc in pybind

youliangtan:
in `rmf_fleet_adapter_python`, remove unnecessary copies of `nlohmann::json desc` objects.

### DIFF
--- a/rmf_fleet_adapter_python/src/adapter.cpp
+++ b/rmf_fleet_adapter_python/src/adapter.cpp
@@ -279,14 +279,12 @@ PYBIND11_MODULE(rmf_adapter, m) {
           [consider_pickup = std::move(consider_pickup)](
             const nlohmann::json &description, Confirmation &confirm)
           {
-            nlohmann::json desc = description;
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter_python/src/adapter.cpp
```diff
@@ -314,8 +311,8 @@ PYBIND11_MODULE(rmf_adapter, m) {
           [consider = std::move(consider)](
             const nlohmann::json &description, Confirmation &confirm)
           {
-            nlohmann::json desc = description;
-            confirm = consider(desc); // confirm is returned by user
+ //           nlohmann::json desc = description;
```
mxgrey, line 314: Nitpick: Let's remove this comment.

mxgrey (changes requested):
Just one very small nitpick

rmf_fleet_adapter_python/src/adapter.cpp
youliangtan: overlooked this, updated!

PR #188, opened 2022-03-18
Update to traffic dependency system

mxgrey:
This PR updates the fleet adapters to use the new traffic dependency system of RMF. Read more on the new system and its benefits [here](https://github.com/open-rmf/rmf_traffic/pull/70).

For the fleet adapters, the API remains *exactly* the same, with just these notes to account for:
* The [original TrafficLight API](https://github.com/open-rmf/rmf_ros2/blob/66721dd2ab5a458c050bad154c6a17d8e4b5c8f4/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/TrafficLight.hpp) is now completely removed. Relative to the new dependency-based system, the old traffic light system was unreliable. Besides it was so difficult to use that I don't believe anyone was using it anyway. The [EasyTrafficLight API](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyTrafficLight.hpp) remains exactly as-is, but should now behave much better since it has been reimplemented to use the new dependency system.
* The "Full Control" API [[1]](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotCommandHandle.hpp) [[2]](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp) remains exactly as-is, except **you no longer need to wait at the time specified by [the plan waypoints](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotCommandHandle.hpp#L72) given to the command handle**. When you receive a new path, your robot can just follow the path all the way to the end. The waypoints given to the command handle will cut off at any waypoint that has a risk of leading into a conflict.

This PR depends on:
* https://github.com/open-rmf/rmf_utils/pull/23
* https://github.com/open-rmf/rmf_traffic/pull/70
* https://github.com/open-rmf/rmf_internal_msgs/pull/37


aaronchongth (approved):
Ran the same test scenarios without seeing any issues so far, this is awesome! LGTM

PR #189, opened 2022-03-25
EasyFullControl API

xiyuoh:
This PR introduces an `EasyFullControl` API for users to initialize robots and set important API functions as callbacks. This provides an easy way of creating their own adapters and also prevents them from having to update the internals of their modified code manually when new changes are pushed.


rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,65 @@
...
+  ///   The API function for navigating your robot to a pose
+  ///   Returns a ProcessCompleted callback to check status of navigation task
+  ///
+  /// \param[in] action_executor
+  ///   The ActionExecutor callback to request the robot to perform an action
+
+  bool add_robot(
+    const std::string& name,
+    const Eigen::Vector3d& pose,
+    std::function<Eigen::Vector3d()> get_position,
```
Yadunund, line 52: Let's alias the type of some of these parameters similar to what you have defined to `ProcessCompleted`. 
Eg.
```
using GetPosition = std::function<Eigen::Vector3d()>;

bool add_robot(
  ...
  GetPosition get_position,
  ...);
 ```

File: rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+  /// \param[in] navigate
+  ///   The API function for navigating your robot to a pose
+  ///   Returns a ProcessCompleted callback to check status of navigation task
+  ///
+  /// \param[in] action_executor
+  ///   The ActionExecutor callback to request the robot to perform an action
+
+  bool add_robot(
+    const std::string& name,
+    const Eigen::Vector3d& pose,
```
Yadunund, line 51: Maybe instead of `Eigen::Vector3d`, this could be a 
```
using Start = std::variant<rmf_traffic::agv::Planner::Start, Eigen::Vector3d>;
...
bool add_robot(
  ...
  Start start
  ...);
 ```
 
 This would allow users to either submit a known `Panner:Start` or a pose. In the latter case, we will use `compute_plan_starts` to estimate the start.

File: rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+    std::function<Eigen::Vector3d()> get_position,
+    std::function<ProcessCompleted(const Eigen::Vector3d pose)> navigate,
+    ActionExecutor action_executor);
+
+private:
+  EasyFullControl();
```
Yadunund, line 57: Use PIMPL pattern

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <rmf_fleet_adapter/agv/RobotUpdateHandle.hpp>
+
+namespace rmf_fleet_adapter {
+namespace agv {
+
+//==============================================================================
+class EasyFullControl : public std::enable_shared_from_this<EasyFullControl>
+{
+public:
+
```
Yadunund, line 31: Add a `static make()` function which will initialize and spin an `Adapter`.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,65 @@
...
+  using ProcessCompleted = std::function<bool(const std::string& id)>;
```
Yadunund, line 32: Since a `ProcessCompleted` is returned for a specific request, we do not need to give it a `string id` to get feedback. The internal implementation will store a map.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,65 @@
...
+  ///
+  /// \param[in] action_executor
+  ///   The ActionExecutor callback to request the robot to perform an action
+
+  bool add_robot(
+    const std::string& name,
+    const Eigen::Vector3d& pose,
+    std::function<Eigen::Vector3d()> get_position,
+    std::function<ProcessCompleted(const Eigen::Vector3d pose)> navigate,
+    ActionExecutor action_executor);
```
Yadunund, line 54: We would also need to accept an object that allows us to perform coordinate transformation between RMF and robot map frames. We could create a `class Transformation` subclass that stores Eigen Transform2D/Affine objects to convert between RMF <> Robot coordinates

Yadunund (commented):
Thanks for getting started on this. Couple of comments for your consideration.

xiyuoh:
@Yadunund Thanks for the comments, addressed them in the latest commit. Also added a `Configuration` class as suggested to store the parameters needed for initializing an adapter and fleet.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,114 @@
...
+  ///   The API function for navigating your robot to a pose.
+  ///   Returns a ProcessCompleted callback to check status of navigation task.
+  ///
+  /// \param[in] action_executor
+  ///   The ActionExecutor callback to request the robot to perform an action.
+  bool add_robot(
+    const std::string& name,
+    Start pose,
+    GetPosition get_position,
+    std::function<ProcessCompleted(const Eigen::Vector3d pose)> navigate,
```
mxgrey, line 100: I'm not super sure about this `navigate` callback. I understand how it would work: We issue a navigation command by triggering this callback, then we get back a new callback that we can poll to figure out if the robot has finished navigating.

What I would've expected is something more like this:

```
std::function<void(const Eigen::Vector3d pose, std::function<void()> notify_finished)> navigate
```

We would trigger this navigate callback and provide the user with:
1. the pose we want the robot to navigate to
2. a callback for the user to trigger when the robot has reached its goal

I do understand the appeal of the current design: The user is less fallible because by polling we can **demand** to know whether the robot is finished yet, whereas giving them a callback to trigger means we're relying on them to remember to trigger it eventually.

However, I think from the user's point of view it will probably be easier for them to just store a callback that they will trigger when their navigating is finished, rather than write a callback that can correctly report whether the destination has been reached whenever it is called. It would also fit better in a reactive programming paradigm.

I don't feel super strongly about this since the current API could absolutely work. Any preference between the two approaches is pretty subjective, so I'd be interested in hearing other people's opinions.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,114 @@
...
+  /// \param[in] name
+  ///   The name of the robot.
+  ///
+  /// \param[in] get_position
+  ///   The position function that returns the robot's current location.
+  ///
+  /// \param[in] navigate
+  ///   The API function for navigating your robot to a pose.
+  ///   Returns a ProcessCompleted callback to check status of navigation task.
+  ///
```
mxgrey, line 93: It seems to me like the `add_robot` function is missing a callback that would let us command the robot to stop immediately. That feature in the [RobotCommandHandle](https://github.com/open-rmf/rmf_ros2/blob/617e83d32f028f2115885dd4f4f6b66799762db6/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotCommandHandle.hpp#L77) does get used sometimes.

mxgrey: I think we're also missing something for [dock](https://github.com/open-rmf/rmf_ros2/blob/617e83d32f028f2115885dd4f4f6b66799762db6/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotCommandHandle.hpp#L87-L89) commands.

xiyuoh: Sorry i didn't mention in this PR that i'm continuing the work on open-rmf's [`feature/easy_full_control`](https://github.com/open-rmf/rmf_ros2/tree/feature/easy_full_control) branch. I added the stop callback there, but let me push some updates from today there as well

xiyuoh:
Closed, continuing on #235

Issue #170, opened 2022-02-11
redesign_v2: missing "completed" status for patrol task with same waypoints


youliangtan:
This is a bug report relevant to redesign v2.

## How to recreate this
**Example1:**

```
ros2 run rmf_demos_tasks dispatch_loop -s coe -f coe -n 1 --use_sim_time
```

websocket printout for `task_state_update`:
```
 
Received [task_state_update] :: Data: 
   {'active': 1, 'assigned_to': {'group': 'tinyRobot', 'name': 'tinyRobot1'}, 'booking': {'id': 'patrol.dispatch-0', 'unix_millis_earliest_start_time': 9229}, 'category': 'Patrol', 'completed': [], 'detail': '', 'estimate_millis': 1974, 'original_estimate_millis': 0, 'pending': [2], 'phases': {}, 'status': 'underway', 'unix_millis_finish_time': 43998, 'unix_millis_start_time': 11577}
 
Received [task_state_update] :: Data: 
   {'active': 1, 'assigned_to': {'group': 'tinyRobot', 'name': 'tinyRobot1'}, 'booking': {'id': 'patrol.dispatch-0', 'unix_millis_earliest_start_time': 9229}, 'category': 'Patrol', 'completed': [], 'detail': '', 'estimate_millis': 1138, 'original_estimate_millis': 0, 'pending': [2], 'phases': {}, 'status': 'underway', 'unix_millis_finish_time': 44127, 'unix_millis_start_time': 11577}
 
Received [task_state_update] :: Data: 
   {'active': 1, 'assigned_to': {'group': 'tinyRobot', 'name': 'tinyRobot1'}, 'booking': {'id': 'patrol.dispatch-0', 'unix_millis_earliest_start_time': 9229}, 'category': 'Patrol', 'completed': [], 'detail': '', 'estimate_millis': 545, 'original_estimate_millis': 0, 'pending': [2], 'phases': {}, 'status': 'underway', 'unix_millis_finish_time': 44580, 'unix_millis_start_time': 11577}
 
Received [task_state_update] :: Data: 
[...]

PR #190, opened 2022-04-05
Add API for reporting status of perform_action

mxgrey:
This PR accomplishes three purposes:
1. The task event status can now be updated by the action executor
2. The action executor can now create event log entries
3. The `perform_action` event status will automatically change to `Completed` when the action executor finishes


mxgrey:
Related to #170

Yadunund (approved):
LGTM!

PR #191, opened 2022-04-07
Add override_status API

Yadunund:
This PR introducing `RobotUpdateHandle::override_status(std::optional<std::string > status)` API.

Users can use this API to override the string status for a given robot. By default the `TaskManger` assigns either `idle` or `working` depending on whether the robot is performing a task or not. But in some cases, users may want to explicitly set the status of the robot. In these scenarios, this API may be used. The API internally checks against the `robot_state.json` schema before accepting the new status. When users want RMF to take back control of the status, they may call this API with `std::nullopt`.
 
### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
@@ -217,19 +217,24 @@ void RobotUpdateHandle::override_status(std::optional<std::string> status)
 
     if (status.has_value())
     {
+      // Here we capture [this] to avoid potential costly copy of
+      // schema_dictionary when more enties are inserted in the future.
+      // It is permissible here since the lambda will only be used within the
+      // scope of this function.
       const auto loader =
-        [n = context->node(), s = _pimpl->schema_dictionary](
+        [context, this](
         const nlohmann::json_uri& id,
         nlohmann::json& value)
         {
-          const auto it = s.find(id.url());
-          if (it == s.end())
+          const auto it = _pimpl->schema_dictionary.find(id.url());
+          if (it == _pimpl->schema_dictionary.end())
           {
             RCLCPP_ERROR(
-              n->get_logger(),
+              context->node()->get_logger(),
               "url: %s not found in schema dictionary. "
-              "Robot status will not be overwritten.",
-              id.url().c_str());
+              "Status for robot [%s] will not be overwritten.",
+              id.url().c_str(),
+              context->name().c_str());
             return;
           }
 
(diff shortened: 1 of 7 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
```diff
@@ -208,6 +208,59 @@ void RobotUpdateHandle::update_battery_soc(const double battery_soc)
...
+          rmf_api_msgs::schemas::robot_state, loader);
+
+        nlohmann::json dummy_msg;
+        dummy_msg["status"] = status.value();
+        validator.validate(dummy_msg);
+
+      }
+      catch (const std::exception& e)
+      {
+        return;
```
mxgrey, line 252: Let's at least log that this problem happened with `RCLCPP_ERROR(...)`, reporting the text of the exception. Since the user is passing in a string, it's very easy for them to have a typo which we should bring to their attention ASAP.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
```diff
@@ -208,6 +208,59 @@ void RobotUpdateHandle::update_battery_soc(const double battery_soc)
...
+void RobotUpdateHandle::override_status(std::optional<std::string> status)
+{
+
+  if (const auto context = _pimpl->get_context())
+  {
+
+    if (status.has_value())
+    {
+      const auto loader =
+        [n = context->node(), s = _pimpl->schema_dictionary](
```
mxgrey, line 221: It's not a huge issue since the schema dictionary only contains one value, but theoretically if it ever increased then this would be a needlessly costly capture, because it makes a complete copy of all the contents of `schema_dictionary`.

This is a very specific case where it's okay to capture `[this]`, because the lambda will only be used within the scope of this function. I recommend just capturing `[this]` here to avoid the unnecessary copy. We should also leave a comment explaining why `[this]` is okay.

mxgrey (changes requested):
Thanks for addressing this need. I'm just requesting a couple small tweaks. Otherwise looks good :+1:

mxgrey (approved):
Thanks for the quick turnaround!

Yadunund:
Sorry @mxgrey I just pushed a commit to update the bindings. Need your review again.

mxgrey (approved):

rmf_fleet_adapter_python/src/adapter.cpp
```diff
@@ -126,6 +126,8 @@ PYBIND11_MODULE(rmf_adapter, m) {
     py::arg("battery_soc"),
     py::call_guard<py::scoped_ostream_redirect,
     py::scoped_estream_redirect>())
+  .def("override_status", &agv::RobotUpdateHandle::override_status,
+    py::arg("battery_soc"))
```
cwrx777, line 130: Is ```py::arg("battery_soc"))``` correct?

mxgrey: Nope, that's definitely a copy/paste error. I'll correct that in an upcoming PR.

Issue #99, opened 2021-08-11
Protect rmf_traffic_schedule node against timeline explosions 


Yadunund:
Its possible due to human errors to not standardize the `use_sim_time` parameter across the `rmf_traffic_schedule` node and different fleet adapters. In such scenarios, when itineraries are set, the schedule database might expand its timeline buckets to cover 51yrs. This explosion causes everything to hang (didn't observe any crashes).

Wondering if we should incorporate a fail-safe against such timeline expansions?

mxgrey:
This is definitely something we should have.

The `Database` class already has a [`cull`](https://github.com/open-rmf/rmf_traffic/blob/40023b0f9b5d79a4781f6105f0368d74c1dfc443/rmf_traffic/include/rmf_traffic/schedule/Database.hpp#L190-L199) function that throws away all trajectories with an end time that predates some timepoint. Maybe the behavior of that function could be beefed up to also filter out any new incoming trajectories that predate the timepoint. That would at least put a lower bound on how "old" the trajectories can be.

An upper bound is much trickier, because if throw away future trajectories we run the risk that the traffic Participant might not know/remember to resend the trajectory when it does become relevant in the future. Maybe an easy way to deal with this, for now, is to add a function to Database that takes in a callback that can be triggered when the Database receives an unreasonably far ahead trajectory. It could look something like this:

```
class Database
{
public:
  /* ... */
  void future_range_warning(
    rmf_traffic::Duration upper_bound,
    std::function<void(std::string details)> callback);
  /* ... */
```

So when a trajectory that exists on the far side of `last_cull_time + upper_bound` is added to the database, `callback` will be triggered with details about who and what is out of bounds. The schedule node can implement `callback` to emit a `RCLCPP_WARN`.

PR #194, opened 2022-04-08
Catch and report bad request change inputs

mxgrey:
It is possible for malformed change requests to come in, so we will catch those situations and report them as errors instead of allowing the node to crash.


youliangtan (approved):
thanks for the quick addition on this catch ex!

Yadunund:
Does this address https://github.com/open-rmf/rmf_ros2/issues/99?

PR #197, opened 2022-04-14
Rmf scheduler ros2

koonpeng:
### Implemented feature

Scheduling system for rmf.

# rmf\_scheduler\_ros2\_client package

Client library for rmf\_scheduler\_ros2.

### Creating a Trigger

```cpp
#include <rclcpp/rclcpp.hpp>

#include <rmf_scheduler_client/payload.hpp>

#include <rmf_scheduler_msgs/srv/create_trigger.hpp>
#include <std_msgs/msg/string.hpp>

#include <chrono>
#include <iostream>

using namespace std::chrono_literals;

int main(int argc, char* argv[])
{
  rclcpp::init(argc, argv);
  auto node = rclcpp::Node::make_shared("example_node");
  auto client = node->create_client<rmf_scheduler_msgs::srv::CreateTrigger>(
    "create_trigger");
  client->wait_for_service(5s);

  std_msgs::msg::String msg;
  msg.data = "hello world";

  auto req =
    std::make_shared<rmf_scheduler_msgs::srv::CreateTrigger::Request>();
  req->trigger.name = "my_trigger";
  req->trigger.at = 100;
  req->trigger.payload = rmf::scheduler::make_serialized_message("test_topic",
      msg);

  auto resp_fut = client->async_send_request(req);
  rclcpp::spin_until_future_complete(node, resp_fut);
  if (resp_fut.get()->success)
  {
    std::cout << "success" << std::endl;
  }
}
```

### Creating a Schedule

Schedules are described using a cron string.

```
┌─────────────seconds (0 - 59)
│ ┌───────────── minute (0 - 59)
│ │ ┌───────────── hour (0 - 23)
│ │ │ ┌───────────── day of the month (1 - 31)
│ │ │ │ ┌───────────── month (1 - 12)
│ │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday)
│ │ │ │ │ │ ┌───────────── years (1970 - 2099) (optional)
│ │ │ │ │ │ │
* * * * * * *
```

| Field | Required | Allowed value | Allowed special characters |
| --- | --- | --- | --- |
| seconds | yes | 0-59 | `*` `,` `-` |
| minutes | yes | 0-59 | `*` `,` `-` |
| hours | yes | 0-23 | `*` `,` `-` |
| days of month | 1-31 | `*` `,` `-` `?` `L` `W` |
| months | yes | 1-12 | `*` `,` `-` |
| days of week | yes | `*` `,` `-` `?` `L` `#` |
| years | no | 1970-2099 | `*` `,` `-` |

The special characters have the following meaning:

| Special character | Meaning | Description |
| --- | --- | --- |
| `*` | all values | selects all values within a field |
| `?` | no specific value | specify one field and leave the other unspecified |
| `-` | range | specify ranges |
| `,` | comma | specify additional values |
| `/` | slash | speficy increments |
| `L` | last | last day of the month or last day of the week |
| `W` | weekday | the weekday nearest to the given day |
| `#` | nth |  specify the Nth day of the month |
Examples: 

| CRON | Description |
| --- | --- |
| * * * * * * | Every second |
| */5 * * * * ? | Every 5 seconds |
| 0 */5 */2 * * ? | Every 5 minutes, every 2 hours |
| 0 */2 */2 ? */2 */2 | Every 2 minutes, every 2 hours, every 2 days of the week, every 2 months |
| 0 15 10 * * ? * | 10:15 AM every day |
| 0 0/5 14 * * ? | Every 5 minutes starting at 2 PM and ending at 2:55 PM, every day |
| 0 10,44 14 ? 3 WED | 2:10 PM and at 2:44 PM every Wednesday of March |
| 0 15 10 ? * MON-FRI | 10:15 AM every Monday, Tuesday, Wednesday, Thursday and Friday |
| 0 15 10 L * ? | 10:15 AM on the last day of every month |
| 0 0 12 1/5 * ? | 12 PM every 5 days every month, starting on the first day of the month |
| 0 11 11 11 11 ? | Every November 11th at 11:11 AM |

Reference: https://github.com/mariusbancila/croncpp/blob/999f7685ab683b58872386c0aa019acf97c6570a/README.md

```cpp
#include <rclcpp/rclcpp.hpp>

#include <rmf_scheduler_client/payload.hpp>

#include <rmf_scheduler_msgs/srv/create_schedule.hpp>
#include <std_msgs/msg/string.hpp>

#include <chrono>
#include <iostream>

using namespace std::chrono_literals;

int main(int argc, char* argv[])
{
  rclcpp::init(argc, argv);
  auto node = rclcpp::Node::make_shared("example_node");
  auto client = node->create_client<rmf_scheduler_msgs::srv::CreateSchedule>(
    "create_schedule");
  client->wait_for_service(5s);

  std_msgs::msg::String msg;
  msg.data = "hello world";

  auto req =
    std::make_shared<rmf_scheduler_msgs::srv::CreateSchedule::Request>();
  req->schedule.name = "my_schedule";
  req->schedule.schedule = "* * * * * *";
  req->schedule.payload = rmf::scheduler::make_serialized_message("test_topic",
      msg);

  auto resp_fut = client->async_send_request(req);
  rclcpp::spin_until_future_complete(node, resp_fut);
  if (resp_fut.get()->success)
  {
    std::cout << "success" << std::endl;
  }
}
```

## Other Endpoints

A list of topics and services exposed by the scheduler node can be found [here](../rmf_scheduler_ros2/README.md). They can use used just like any other ros2 endpoints.

## Quality Declaration

This package claims to be in the **Quality Level 4** category. See the [Quality Declaration](QUALITY_DECLARATION.md) for more details.


rmf_scheduler_ros2/src/Publisher.cpp
```diff
@@ -0,0 +1,82 @@
...
+  auto& rcl_inner =
+    inner_sermsg.get_rcl_serialized_message();
+  std::copy(rmf_msg.data.data(),
+    rmf_msg.data.data() + rmf_msg.data.size(), rcl_inner.buffer);
+  rcl_inner.buffer_length = rmf_msg.data.size();
+
+  auto pub = this->node->create_generic_publisher(rmf_msg.topic_name,
+      rmf_msg.message_type,
+      rclcpp::SystemDefaultsQoS{});
+  pub->publish(inner_sermsg);
```
luca-della-vedova, line 79: Gave it a quick spin and looking good so far!
I have a mix of feedback and question on reliability since I see that the approach is to create a publisher, publish a message then destroy it immediately.

I observed that "sometimes" some of my subscribers don't get the messages even though they are subscribing to the topic being published. _I believe_ this might be a discovery issue, since the publisher creation and subscription immediately follow one another it might happen that all the subscribers were not discovered.
I tested this case by running a few subscribers in parallel (i.e. a few applications, a `ros2 topic echo`) and adding a trigger to be ran as soon as possible (in the past), together with a trigger to run a bit in the future. Sometimes one of my applications (really only the `ros2 topic echo`) misses the first message, but seems to always work for the second message (maybe due to some caching happening at the ros2 layer for subscribers?).

Adding a very hacky `usleep(1e6)` between publisher creation and message publishing worked but that is most likely not a great way to deal with this

koonpeng: Hmm I'm not sure how to go about addressing this. The node is created early and stays alive throughout so I thought this won't be an issue, at least it works on the e2e test :thinking:. Can you try the e2e tests and see if it works on your pc? Even better, it would be really helpful if we can reproduce it in the tests.

I'm thinking a possible solution is to wait for subscribe count > 0 before publishing, but that may only work when considering only 1 subscriber, when there are more subscribers we don't know how many subscribers to be discovered before publishing.

We can keep the publisher alive but I'm not sure if it will work because afaik discovery is based on nodes and not publishers. Even if it works, it wouldn't be a good solution as we will end up with a publisher for each trigger/schedule, possibly resulting in hundreds or maybe even thousands of publishers.

luca-della-vedova: The E2E tests ran through `colcon test --packages-select rmf_scheduler_ros2_client` all fail on my machine and I'm not quite sure why:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="e2e_TEST" errors="0" failures="5" tests="5" hostname="tbd" time="6.547" timestamp="2022-04-21T09:33:40Z">
    <testcase classname="e2e_TEST.global" name="create trigger" time="0.000" status="run">
      <failure type="FAIL">
FAILED:
Timed out waiting for new trigger to be executed
at /home/luca/ws_avcs/src/rmf/rmf_ros2/rmf_scheduler_ros2_client/test/e2e_TEST.cpp:139
      </failure>
    </testcase>
    <testcase classname="e2e_TEST.global" name="cancel_schedule" time="0.000" status="run">
      <failure type="FAIL">
FAILED:
Timed out waiting for response
at /home/luca/ws_avcs/src/rmf/rmf_ros2/rmf_scheduler_ros2_client/test/e2e_TEST.cpp:79
      </failure>
    </testcase>
    <testcase classname="e2e_TEST.global" name="list schedules" time="0.000" status="run">
      <failure type="FAIL">
FAILED:
Timed out waiting for response
at /home/luca/ws_avcs/src/rmf/rmf_ros2/rmf_scheduler_ros2_client/test/e2e_TEST.cpp:79
      </failure>
    </testcase>
    <testcase classname="e2e_TEST.global" name="list schedule states" time="0.000" status="run">
      <failure type="FAIL">
FAILED:
Timed out waiting for response
at /home/luca/ws_avcs/src/rmf/rmf_ros2/rmf_scheduler_ros2_client/test/e2e_TEST.cpp:79
      </failure>
    </testcase>
    <testcase classname="e2e_TEST.global" name="cancel all" time="0.000" status="run">
      <failure type="FAIL">
FAILED:
Timed out waiting for response
at /home/luca/ws_avcs/src/rmf/rmf_ros2/rmf_scheduler_ros2_client/test/e2e_TEST.cpp:79
      </failure>
    </testcase>
    <system-out/>
    <system-err/>
  </testsuite>
</testsuites>
```
I even get `uncrustify` errors:

```
luca@luxury:~/ws_avcs$ colcon test-result
build/rmf_scheduler_ros2_client/Testing/20220421-0935/Test.xml: 2 tests, 0 errors, 2 failures, 0 skipped
build/rmf_scheduler_ros2_client/test_results/rmf_scheduler_ros2_client/e2e_TEST.catch2.xml: 5 tests, 0 errors, 5 failures, 0 skipped
build/rmf_scheduler_ros2_client/test_results/rmf_scheduler_ros2_client/uncrustify.xunit.xml: 2 tests, 0 errors, 1 failure, 0 skipped

Summary: 65100 tests, 0 errors, 8 failures, 0 skipped
```

I also find it a bit suspicious that CI is green, I can't figure out how building can pass if the dependency PR under rmf_internal_msgs is not merged, shouldn't compilation fail because it fails to find the message headers?

luca-della-vedova: > I'm thinking a possible solution is to wait for subscribe count > 0 before publishing, but that may only work when considering only 1 subscriber, when there are more subscribers we don't know how many subscribers to be discovered before publishing.

I think this is an OK solution, I am a bit worried of the edge case of having no subscribers and what would happen, I guess an "async wait with a timeout for num_subscribers > 0" should cover most cases?

koonpeng: I wouldn't depend on this if possible, this solution doesn't really work because we can't know how many subscribers there will be, if we published as soon as there is 1 subscriber, the 2nd subscriber will not receive the message, leading to the same problem again.

A better solution is to have subscribers not rely on one off messages during its startup sequence or fix the launch system to ensure targets are properly started in order.

rmf_scheduler_ros2_client/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+  ${rmf_scheduler_msgs_INCLUDE_DIRS}
+)
+
+#===============================================================================
+
+if(BUILD_TESTING)
+  find_package(ament_cmake_catch2 REQUIRED)
+  find_package(ament_cmake_uncrustify REQUIRED)
+  find_package(rmf_utils REQUIRED)
+  find_package(Boost REQUIRED COMPONENTS filesystem)
```
luca-della-vedova, line 33: Just a nit, on docker builds this dependency is bringing in all of the boost library, I saw that in the other package std::filesystem is being used instead, any chance to also use it here?

koonpeng: iirc, this is needed for the `boost::process`, it doesn't allow using paths from `std::filesystem`.

luca-della-vedova:
Package has been moved to a [new repo](https://github.com/open-rmf/rmf_scheduler). Closing the PR

PR #198, opened 2022-04-15
Allow system integrators to specify an update listener

mxgrey:
As an alternative to (or in combination with) the websocket publishing of updates, this PR allows system integrators to provide their own callback to listen to update messages. System integrators can then do whatever they wish with those messages.


aaronchongth (approved):
Thanks for adding this API, this will be so useful for upcoming integrations!

PR #199, opened 2022-04-18
Add a WaitUntil event and use it for ResponsiveWait

mxgrey:
This PR adds a `WaitUntil` event and uses it at the end of each `ResponsiveWait` cycle. This will relieve system integrators of the need to wait until the last waypoint's time before triggering the `path_finished_callback` of `RobotCommandHandle::follow_new_path`. Now system integrators can just trigger that callback as soon as the robot arrives at the end of the path.


aaronchongth (dismissed):
Thanks for adding this so quickly! LGTM

PR #200, opened 2022-04-19
Publish canceled status after cancel queued task

youliangtan:
This fix addresses the issue where `rmf-web` dashboard will not update the status of a "queued" task as "canceled" after a successful cancellation. Publish a "canceled" status after removing the task from the queue.


youliangtan:
Close this, as this fix is done in https://github.com/open-rmf/rmf_ros2/pull/205

PR #202, opened 2022-04-23
No exception for plan_id

mxgrey:
Related to https://github.com/open-rmf/rmf_traffic/pull/76 this PR will now check if the `plan_id` was accepted, notify if it was not accepted, and take drastic action if the rejections persist.


youliangtan (approved):
Thanks for the quick change!

PR #205, opened 2022-05-05
Add lookahead behavior for GoToPlace and fix various sources of seg faults

mxgrey:
This PR came from some rapid development cycles while getting a demo to work, so it's a mixed bag of changes.

* Adds a traffic schedule lookahead ability for `GoToPlace`, which helps to smooth out traffic negotiations for tasks that involve passing through a sequence of waypoints (e.g. patrolling)
* Removes stdout/stderr redirect for the python bindings, since that was found to cause segfaults (this is a known issue which should be fixed in newer versions of pybind11: https://github.com/pybind/pybind11/pull/2675)
* Fixes a logic error for the cancel + kill API which prevented those APIs from working unless the user provides a callback
* Provides the most general possible `update_position` overload to give system integrators maximum flexibility in how they report their robot's position


rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -1678,49 +1736,130 @@ void TaskManager::_publish_task_state()
...
 
-    pending_json["category"] = info.category;
-    pending_json["detail"] = info.detail;
+  pending_json["unix_millis_start_time"] =
+    to_millis(pending.deployment_time().time_since_epoch()).count();
 
-    pending_json["unix_millis_start_time"] =
-      to_millis(pending.deployment_time().time_since_epoch()).count();
+  copy_assignment(pending_json["assigned_to"], *_context);
+  pending_json["status"] = "canceled";
```
aaronchongth, line 1814: Got confused by this for a while and thought it was a typo, we might need to converge to single or double L at some point, I found multiple locations where we use double Ls, https://github.com/open-rmf/rmf_api_msgs/blob/main/rmf_api_msgs/schemas/task_state.json#L51.

This should not block merging this PR though

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Node.cpp
```diff
@@ -92,17 +92,22 @@ std::vector<ScheduleNode::ConflictSet> get_conflicts(
...
         const auto* dep_u =
           route->check_dependencies(vc->participant, vc->plan_id, vc->route_id);
-        if (dep_u && !dep_u->empty())
+        if (dep_u)
           continue;
 
+//        std::cout << "No deps for " << description->name()
+//                  << " " << plan_id << "|" << r << " vs "
+//                  << vc->description.name() << " " << vc->plan_id << "|" << vc->route_id
+//                  << std::endl;
```
aaronchongth, line 106: Are these comments still useful?

aaronchongth (approved):
These features look great! The refactoring in Task manager also makes it easier to go through, thanks!

I just have 2 pretty trivial comments, but otherwise LGTM! Should we be updating the change logs for these packages as well? You can also mention the newly added API for cancelling or killing tasks from `RobotUpdateHandle`, those were extremely useful :sparkles:

aaronchongth (changes requested):
On second thought, since these are using features in `2.1.0` of `rmf_task` and `rmf_task_sequence`, I'd say we should probably consider bumping a minor release to match as well. I'm not super well versed in our versioning workflow, let me know what you think.

youliangtan:
I tested a few scenarios, with queue task cancellation and the new generic `update_position()` API for some demo use cases. All works perfectly :+1:

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
mxgrey: Yeah, the inconsistency between the number of `l`s in `canceled` versus `cancellation` probably creates a high risk of typos, but I chose those spellings as they're the more common spellings in American English, which is usually preferred over British spellings in the global software industry. I'm open to revising the API if people anticipate that it will cause problems.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Node.cpp
mxgrey: Thanks for catching that, comments are removed in [9433066](https://github.com/open-rmf/rmf_ros2/pull/205/commits/9433066fd92c2d53d5c3025e8bc88efe36f97bea)

mxgrey:
@aaronchongth great idea, CHANGELOG is updated in [46d1cac](https://github.com/open-rmf/rmf_ros2/pull/205/commits/46d1cac47f0a9c2c9b047d24e674336faca4ad4d). Let me know if there's anything else to block this PR.

aaronchongth (approved):
Thanks for the changes fixes! 🙇LGTM!

PR #206, opened 2022-05-06
state as underway during perform action

youliangtan:
During performaction event, the state was shown as "standby". It is more understandable to users if the state is shown as "underway" during an ongoing perform action task. Single line fix
.


Yadunund:
Btw @youliangtan you can also use the override API to set the status from the fleet adapter implementation
https://github.com/open-rmf/rmf_ros2/pull/191

youliangtan:
The override api is to update the robot status. The updated underway status is a task status.

Yadunund:
Thanks for clarifying. I got the two mixed up for a moment 😅

PR #207, opened 2022-05-11
Fleet adapter publishes its navigation graph

Yadunund:
This PR makes the fleet adapter publish its navigation graph as a `building_map_msgs/Graph` msg over the `/nav_graphs` topic. The QoS is set to durability=transient_local, reliability=reliable. 

This is meant to be a static graph. For each waypoint, properties such as `map_name`, `is_holding_point`, `is_charger` etc are appended to the `Params[]` field in `GraphNode`.

> Note: The status of each lane, ie, whether open or closed along with lane properties like speed limits will be published over a separate topic. This will be implemented in a separate PR.

On a side note, I tried using the `message::build` pattern as seen below
```
rmf_building_map_msgs::msg::GraphNode node;
rmf_building_map_msgs::msg::Param param;
...
...
node.params.emplace_back(
    rmf_building_map_msgs::build<rmf_building_map_msgs::msg::Param>()
    .name("is_holding_point")
    .type(param.TYPE_BOOL)
    .value_int(0)
    .value_float(0.0)
    .value_string("")
    .value_bool(false)
);
```
but the compiler throws the error:
```
error: use of ‘auto rmf_building_map_msgs::build() [with MessageType = rmf_building_map_msgs::msg::Param_<std::allocator<void> >]’ before deduction of ‘auto’

```
Out of curiosity, what was I doing wrong?


mxgrey:
> Out of curiosity, what was I doing wrong?

Try using `.push_back(...)` instead of `.emplace_back(...)`. The function `emplace_back` does "perfect forwarding" which means it needs to infer the type that you're meaning to pass in. Sometimes that doesn't work great. I'm not entirely sure if that's actually the issue because I've never seen that particular error before, but that would be my best guess.

mxgrey:
I think adding reusable serialization and deserialization functions to a library is a great idea, so big :+1: to that.

Adding those functions to `rmf_traffic_ros2` sounds perfect to me as well.

I took a moment to consider whether the graph message belongs in `rmf_traffic_msgs` instead of `rmf_building_map_msgs` so that there's better symmetry between the `rmf_traffic` objects and their `rmf_traffic_msgs` serializations, but I haven't come to any strong conclusions on that. On one hand, it makes sense for the graph message to go into `rmf_traffic_msgs` because it's the serialization of a data structure that's from `rmf_traffic`. On the other hand, virtually everything else in `rmf_traffic_msgs` is specific to the traffic scheduling and negotiation system, which is not really the case for the navigation graph.

In conclusion, I see nothing wrong with keeping the graph message in `rmf_building_map_msgs`, especially to minimize the disruptiveness of the change.

Yadunund:
Thanks for the quick feedback!

> In conclusion, I see nothing wrong with keeping the graph message in rmf_building_map_msgs, especially to minimize the disruptiveness of the change.
Roger 👍🏼

Yadunund:
@mxgrey I've updated the implementation as discussed. 

It's not apparent to me how we can serialise an`Lane::EventPtr` to determine whether there is a door/lift open/close event or dock/wait event. 
Similarly with OrientationConstraints. So I've left a note in the API documentation highlighting this limitation.

luca-della-vedova:
Just a high level comment since I worked a bit on the linked [graph conversion function](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_traffic_ros2/include/rmf_traffic_ros2/agv/Graph.hpp).
Something that I was worried about when doing that implementation was code duplication with the existing [parse_graph](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/parse_graph.cpp). While it could not really be helped in that case (we were parsing a brand new file format, that doesn't have vertex indexes, only works in WGS84 coordinates, doesn't really support multilevel buildings) I feel it could be avoided in this case.

From what I understand the function under parse_graph parses a string which is a yaml navgraph and converts it into a `rmf_traffic::agv::Graph` while the function you wrote does bidirectional conversion between the `rmf_building_map_msgs::Graph` and the `rmf_traffic::agv::Graph`.
I feel there aren't huge differences between parsing a raw yaml and the rmf_building_map_msgs and now we have yet another graph parser, yaml, rmf_site_map_msgs and rmf_building_map_msgs with almost (but not full) feature parity, it's not great to have to change three separate parsers whenever a change in file format is introduced (i.e. a new property to the graph)

* Would it be possible to deprecate at least one? A possible idea could be that the fleet adapter subscribes to the whole building map, finds its own graph and republishes it for downstream users, in this way we can deprecate `parse_graph` and only use the new `rmf_building_map_msgs` parsers.
* An alternative, especially if you want the publishing to be static and done once at running time, could be to have this done by the building_map_server but I guess the idea is that fleet adapters are in charge of their own navgraph?
* Another alternative could be to have an approach similar to what is done for the `rmf_site_map_msgs`, where the whole map is just published as a string, then if parsing (i.e. `parse_graph`) was part of a shared library it can be used among all nodes that need to process it.

My personal favorite is the first option, but of course I'm biased because I was an advocate of the dynamic navgraph approach in the past :sweat_smile: .
The building map server already publishes a navgraph when it publishes a map, specifically for each [level](https://github.com/open-rmf/rmf_building_map_msgs/blob/main/rmf_building_map_msgs/msg/Level.msg#L6) it publishes a [navgraph](https://github.com/open-rmf/rmf_traffic_editor/blob/main/rmf_building_map_tools/building_map_server/building_map_server.py#L190-L194), it could require a bit of tuning to make sure all the properties are there (for example on a quick glance I couldn't see support for multiple coordinate systems) but the code path to generate the message and the navgraph yaml is the same (they both call this [function](https://github.com/open-rmf/rmf_traffic_editor/blob/main/rmf_building_map_tools/building_map/level.py#L428)) so most features should be there.
Now since the properties in the message served by building_map_server and the properties in the static yaml are the same, we should be able to pretty much copy the whole parsing logic from parse_graph (and write the serialization) to have full feature parity, including orientation constraints, doors and lift events.

The biggest drawback is that now the map server has to be running for fleet adapters to be initialized which is an additional dependency and I don't know what the implications might be. On the other hand, we would be able to get rid of the dependency in fleet adapters for nav graph files (i.e. [here](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/launch/fleet_adapter.launch.xml#L12)) which would now become just an index and I can envision in the future we could have a naming system in graphs so that instead of using 0-9 integers we use the fleet name, that would get rid of the navgraph parameter altogether.

What do you think?

Yadunund:
@luca-della-vedova  I think you've misunderstood the functionality here. We're not converting between the `rmf_building_map_msgs::BuildingMap` message published by the `rmf_building_map_server` and the `rmf_traffic::agv::Graph` object.

We're adding a utility to publish the exact navigation graph used by the fleet adapter for its planning. For convenience, we're re-using the `rmf_building_map_msgs::Graph` message definition. 

> The biggest drawback is that now the map server has to be running for fleet adapters to be initialized which is an additional dependency and I don't know what the implications might be.

There is no subscription anywhere in this implementation to the building map server and hence no dependency.

We've had this discussion before and we agreed that it is best if the fleet adapters are the ones to publish the exact nav graph they are using. The building map server has no way to know whether a fleet is actually configured to use a nav graph of certain index. Even if we update some definitions to map the index to the fleet  name, there is no guarantee that the fleet adapter was actually loaded with this navgraph. Maybe they don't even use the exported nav_graph.yaml.

Yadunund:
Commit https://github.com/open-rmf/rmf_ros2/pull/207/commits/f936da7b3d00011700e48052391e023e28411d89 ensures that all lane events (DoorOpen, DoorClose, LiftSessionBegin, LiftDoorOpen, LiftMove, LiftSessionEnd) and orientation constraints are serialized/deserialized.

@mxgrey this feature is complete and ready for review.

rmf_traffic_ros2/src/rmf_traffic_ros2/convert_Graph.cpp
```diff
@@ -325,7 +340,7 @@ rmf_traffic::agv::Graph json_to_graph(
     }
 
     // dock_name is only applied to the lane going to the waypoint, not exiting
-    const rmf_traffic::Duration duration = std::chrono::seconds(5);
+    const rmf_traffic::Duration duration = std::chrono::nanoseconds(5);
```
mxgrey, line 343: I'm sure the choice of 5 second docking times was an arbitrary default, but I think 5ns is even more arbitrary. At that point we may as well make the default 0ns.

@luca-della-vedova do you have anything that depends on a 5 second default duration for docking? If not, I suggest we change the default to 0s.

mxgrey (approved):
I think it's fine to merge, but I do have one remark in an inline comment.

rmf_traffic_ros2/src/rmf_traffic_ros2/convert_Graph.cpp
luca-della-vedova: No problem for my side! The 5 seconds was taken from [parse_graph](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/parse_graph.cpp#L286) so I guess to keep the implementation consistent all of them should have the same default?

Yadunund: Whoops that diff is an unintended consequence of search and replace. I can revert it back to 5 seconds?

Yadunund:
Could I get a re-approval for this?

luca-della-vedova (approved):

PR #212, opened 2022-06-07
ws broadcast client in dispatcher node

youliangtan:
Currently, if a task bidding failed, no task state update will get broadcasted to the ws server. From the rmf-web dashboard, this failed task will get shown as "unknown".

 This solution simply implements the same`BroadcastClient` in the dispatcher_node. Similarly, the user can specify the broadcast uri via rosparam `server_uri`.

Since the impl is similar to the one in `rmf_fleet_adapter` pkg, wonder if it is preferred to create a common lib for task json handling and ws `BroadcastClient`.

### DIFF
--- a/rmf_fleet_adapter/schemas/event_description__go_to_place.json
+++ b/rmf_fleet_adapter/schemas/event_description__go_to_place.json
@@ -16,7 +16,8 @@
           "type": "array",
           "items": { "$ref": "place.json" }
         }
-      }
+      },
+      "required": ["place"]
     }
   ]
 }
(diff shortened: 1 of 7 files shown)


mxgrey:
Both implementations exist within `rmf_ros2`. I'd recommend creating a bite-sized package just to hold the broadcast client. We can call the new package `rmf_websocket` or something to that effect.

youliangtan:
I created the `rmf_websocket` pkg, with `BroadcastClient` and `BroadcastServer` classes as public apis. Let me know if they make sense. 

Side note, I'm not able to replicate the failed CI tests locally. Will soon investigate this odd stuff further.

mxgrey (approved):
Thanks for going through the trouble of refactoring! The CI seems to be passing after I made [this tweak](https://github.com/open-rmf/rmf_ros2/pull/212/commits/2ea6310ef5b8a07f706fc34043b5cb657930dc66). I don't know why the error was happening in the CI but not on our local machines, but :shrug: The fix is important for robustness anyway.

PR #215, opened 2022-06-13
Changes for humble compatibility

luca-della-vedova:
### Fixed bug

This PR fixes compile errors for rmf_ros2 under humble.

### Fix applied

Specifically, issues with `optional` not being found in bidding Reponse [here](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_task_ros2/include/rmf_task_ros2/bidding/Response.hpp#L41), as well as an error with the Python bindings:

```
CMake Error at /usr/lib/cmake/pybind11/pybind11NewTools.cmake:207 (python3_add_library):
  Unknown CMake command "python3_add_library".
Call Stack (most recent call first):
  CMakeLists.txt:38 (pybind11_add_module)
```

According to the [CMake documentation](https://cmake.org/cmake/help/latest/module/FindPython3.html), we need to add a find_package(Python3) to add the CMake command.

### DIFF
--- a/rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/YamlSerialization.cpp
+++ b/rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/YamlSerialization.cpp
@@ -85,7 +85,7 @@ rmf_traffic_msgs::msg::ConvexShape convex_shape(YAML::Node node)
 
   rmf_traffic_msgs::msg::ConvexShape shape;
   shape.type = shape_type(node[TypeKey]);
-  shape.index = node[IndexKey].as<uint8_t>();
+  shape.index = node[IndexKey].as<uint16_t>();
   return shape;
 }
 
@@ -249,7 +249,7 @@ YAML::Node serialize(rmf_traffic_msgs::msg::ConvexShape shape)
 {
   YAML::Node node;
   node[TypeKey] = serialize_shape_type(shape.type);
-  node[IndexKey] = shape.index;
+  node[IndexKey] = static_cast<uint16_t>(shape.index);
   return node;
 }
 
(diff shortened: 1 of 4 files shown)


luca-della-vedova:
Also added an upcast from uint8_t to uint16_t for the schedule yaml database.

It seems that yaml-cpp [has trouble](https://github.com/jbeder/yaml-cpp/issues/1081) dealing with uint8_t values, specifically we were getting "bad access" exceptions and the node was failing to initialize.
It is not a super ideal fix but we upcast to uint16_t to write to the yaml then there will be a narrowing conversion when parsing it into a uint8_t variable on reading.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/YamlSerialization.cpp
```diff
@@ -85,7 +85,7 @@ rmf_traffic_msgs::msg::ConvexShape convex_shape(YAML::Node node)
 
   rmf_traffic_msgs::msg::ConvexShape shape;
   shape.type = shape_type(node[TypeKey]);
-  shape.index = node[IndexKey].as<uint8_t>();
+  shape.index = node[IndexKey].as<uint16_t>();
```
mxgrey, line 88: The [message definition](https://github.com/open-rmf/rmf_internal_msgs/blob/c0b929db6ca666f7951dc633c573892a7c812c28/rmf_traffic_msgs/msg/ConvexShape.msg#L13) has this field as a `uint8`. Is there a reason to change the type to `uint16_t` here?

mxgrey (commented):
Generally looks good, except I don't understand the motive behind the integer type change. Did yaml-cpp stop supporting `uint8_t`?

PR #219, opened 2022-08-02
Fix rmf_obstacle_ros2 intersection checker

PranayShirodkar:
## Summary
test_IntersectionChecker.cpp failing Test 3.

IntersectionChecker.cpp provides methods to check if 2 bounding boxes are intersecting. If they are not intersecting, it returns the distance between the 2 bounding boxes.

## Changes

1. `make_transformed_box(...)` method was fixed in IntersectionChecker.cpp
2. Expected distance between bounding boxes was updated in test_IntersectionChecker.cpp
3. Test 4 & 5 created to check when both bounding boxes are at random positions and orientations

## Result

1. Test 3 is passing
2. Test 1 & 2 passing with smaller error margin
3. Test 4 & 5 passing

Onshape was used to visualize the test cases and measure the expected distance between non-intersecting bounding boxes.
https://cad.onshape.com/documents/437b8a87624f92ed41e30c3b/w/fd8e8a03d1faf030c1f83614/e/7bbd8a761aa9f5c4f8223de4?renderMode=0&uiState=62e8da33b51e8d734cd12865

### DIFF
--- a/rmf_obstacle_ros2/src/lane_blocker/LaneBlocker.cpp
+++ b/rmf_obstacle_ros2/src/lane_blocker/LaneBlocker.cpp
@@ -26,7 +26,7 @@
 #include <geometry_msgs/msg/pose_stamped.hpp>
 #include <geometry_msgs/msg/vector3_stamped.hpp>
 
-#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
 
 #include <rclcpp_components/register_node_macro.hpp>
 
--- a/rmf_obstacle_ros2/src/lane_blocker/LaneBlocker.cpp
+++ b/rmf_obstacle_ros2/src/lane_blocker/LaneBlocker.cpp
@@ -26,7 +26,7 @@
 #include <geometry_msgs/msg/pose_stamped.hpp>
 #include <geometry_msgs/msg/vector3_stamped.hpp>
 
-#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
 #include <rclcpp_components/register_node_macro.hpp>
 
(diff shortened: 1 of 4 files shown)


rmf_obstacle_ros2/src/lane_blocker/LaneBlocker.cpp
```diff
@@ -26,7 +26,7 @@
 #include <geometry_msgs/msg/pose_stamped.hpp>
 #include <geometry_msgs/msg/vector3_stamped.hpp>
 
-#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
```
Yadunund, line 29: Please undo this change. `Galactic` version of `tf2_geometry_msgs` only contains `.h` header while `Humble` contains both. So until we move to I turtle, the original is fine and will enable us to build on both `Galactic` and `Humble`.

rmf_obstacle_ros2/src/lane_blocker/test/test_IntersectionChecker.cpp
```diff
@@ -39,6 +39,7 @@ SCENARIO("Test IntersectionChecker")
   WHEN("AABB geometries are not intersecting and 1m apart")
   {
     double how_much;
+    double expected = 1.0;
```
Yadunund, line 42:
```suggestion
    const double expected = 1.0;
```

Same for all below.

Yadunund (changes requested):
Thanks for fixing this bug! Everything works great but I have a couple of small issues I've raised.

rmf_obstacle_ros2/src/lane_blocker/LaneBlocker.cpp
PranayShirodkar: my bad, i didnt realize that!

rmf_obstacle_ros2/src/lane_blocker/test/test_IntersectionChecker.cpp
PranayShirodkar: updated this, thanks!

Yadunund (approved):
Thanks for addressing the feedback!

PR #210, opened 2022-05-18
rmf_obstacle_ros2

Yadunund:
This PR introduces the `rmf_obstacle_ros2` package. It depends on https://github.com/open-rmf/rmf_internal_msgs/pull/42

The package currently provides:
1. `rmf_obstacle_ros2` library that provides
    a. `Detector` and `Responder` pure abstract classes that users may implement to provide detection results and take action on these results respectively. The derived classes should be implemented as dynamic plugins that can be loaded by `pluginlib`.
    b. `Obstacle` concrete class to store 3D representation of obstacles. Also includes utility functions to convert between `rmf_obstacle_ros2::Obstacle` and `rmf_obstacle_msgs::ObstacleData` thereby serializing/deserializing 3D octrees.
 
2. `obstacle_manager` that dynamically loads implemented `Detector` and `Responder` instances and publishes detected obstacles (Obstacles.msg) to `/rmf_obstacles` and optionally takes action on the detection results. The `detector_plugin` and `responder_plugin` ROS 2 params need to be specified with the fully qualified names of the implemented plugins. 

For example
``` ros2 run rmf_obstacle_ros2 obstacle_manager --ros-args -p detector_plugin:=my_plugin::MyDetector -p responder_plugin:=my_plugin::MyResponder``` 

> Note: If no `responder_plugin` is specified, the node will only publish detection results to `/rmf_obstacles`.

PR #222, opened 2022-08-16
Add Speed Limit Requests

PranayShirodkar:
### continuous_checker

- Add continuous_checker param to LaneBlocker node. This param is a bool with default value false.
- If obstacles are static, continuous_checker should be left as false. If obstacles are dynamic/moving, set continuous_checker to true.
- When continuous_checker is false, the LaneBlocker node will only check an obstacle for intersection with all lanes when the obstacle first appears.
- When continuous_checker is true, the LaneBlocker node will continuously check an obstacle for intersection with all lanes as long as the obstacle is present.

### add speed limits

- Lanes can now be in Closed state, Speed Limited state and normal state
- Refactored the lane open and lane close msg creation and sending
- Add speed limit and speed unlimit msg
- Handle all 6 possible state transitions
- Add python bindings for agv::FleetUpdateHandle::limit_lane_speeds() and agv::FleetUpdateHandle::remove_speed_limits()  - related PR in [rmf_demos](https://github.com/open-rmf/rmf_demos/pull/152)


mxgrey:
@Yadunund since this is targeting your draft PR #210 I'll leave it up to you if/when/how this gets merged.

Yadunund (approved):
I've left TODOs for things to address before merging into `main`. Fixed const correctness and made an API tweak to the publisher functions to prevent segfaults if the unordered_maps containing the unique_ptrs are accessed after moving the messaging during `publish()`.

PR #223, opened 2022-08-18
create unique task_id with timestamp

youliangtan:
The current way of generating a task id uses a `task_counter`, which generates the task_id by appending an integer `task counter` value (eg. `patrol.dispatch-1`). However, this poses a problem when we require the task_id to be unique even after restarting the `task dispatcher` node (during deployment setup especially). Therefore, this solution generates a unique id with a timestamp. This can be enabled via ros2 param, by setting  `use_timestamp_for_task_id` as `true`. 

p/s: `use_timestamp_for_task_id` is false by default (using task counter)

## To test

Run dispatcher with use_timestamp_for_task_id as true
```bash
ros2 run rmf_task_ros2 rmf_task_dispatcher  --ros-args -p use_timestamp_for_task_id:=true
```

Try dispatch a dummy task
```bash
ros2 run rmf_demos_tasks dispatch_patrol -p coe
```

### DIFF
--- a/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
+++ b/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
@@ -486,7 +486,7 @@ public:
       if (use_timestamp_for_task_id)
       {
         task_id += std::to_string(
-          static_cast<int>(node->get_clock()->now().nanoseconds()/1e9));
+          static_cast<int>(node->get_clock()->now().nanoseconds()/1e6));
       }
       else
       {
--- a/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
+++ b/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
@@ -175,6 +175,7 @@ public:
   builtin_interfaces::msg::Duration bidding_time_window;
   std::size_t terminated_tasks_max_size;
   int publish_active_tasks_period;
+  bool use_timestamp_for_task_id;
 
   std::unordered_map<std::size_t, std::string> legacy_task_type_names =
   {
@@ -209,7 +210,11 @@ public:
     RCLCPP_INFO(node->get_logger(),
       " Declared publish_active_tasks_period as: %d secs",
       publish_active_tasks_period);
-
+    use_timestamp_for_task_id =
+      node->declare_parameter<bool>("use_timestamp_for_task_id", false);
+    RCLCPP_INFO(node->get_logger(),
+      " Use timestamp with task_id: %s",
+      (use_timestamp_for_task_id ? "true" : "false"));
 
     std::optional<std::string> server_uri = std::nullopt;
     const std::string uri =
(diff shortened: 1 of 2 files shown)


Yadunund (approved):
LGTM. Can you update the CHANGELOG?

rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
```diff
@@ -474,9 +479,20 @@ class Dispatcher::Implementation
...
-      const std::string task_id =
+      std::string task_id =
         task_request_json["category"].get<std::string>()
-        + ".dispatch-" + std::to_string(task_counter++);
+        + ".dispatch-";
+
+      if (use_timestamp_for_task_id)
+      {
+        task_id += std::to_string(
+          std::chrono::duration_cast<std::chrono::seconds>(
```
Yadunund, line 489: Actually, why do we need the duration_cast? `time_since_epoch().count()` would directly return the nanoseconds since the start.
Also, should we use node time here? (`node->get_clock()->now()`)

rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
```diff
@@ -474,9 +479,19 @@ class Dispatcher::Implementation
...
-      const std::string task_id =
+      std::string task_id =
         task_request_json["category"].get<std::string>()
-        + ".dispatch-" + std::to_string(task_counter++);
+        + ".dispatch-";
+
+      if (use_timestamp_for_task_id)
+      {
+        task_id += std::to_string(
+          static_cast<int>(node->get_clock()->now().nanoseconds()/1e9));
```
Yadunund, line 489: Wondering if we should leave the timestamp in nanoseconds and not divide by 1e9 to get seconds? If somehow two or more tasks get processed within the same second, they might have the same task_id?

youliangtan: Yea, that is valid concern that i thought about too. The only concern (probably insignificant) is having nanosec as timestamp is way too long. This will result in a long `task_id` that might look quite ugly on the dashboard? Example shown here:

Screenshot from 2022-08-19 15-40-39

Let me nego abit, how about milisec? Since  `unix_millis` time is regularly used when defining our rmf time field.

Yadunund: Okay let's go with millis for now.
I do this this feature will be redundant once we implement a proper backup and restore functionality for this node. That way when the node is restored after a crash, it will know the last issued task_id.

Yadunund (approved):
Thanks for the addressing the feedback.

Any idea why CI is failing?
image

youliangtan:
Thanks to the approval and addressing the failed CI https://github.com/open-rmf/rmf_ros2/pull/225 . Will merge this in since the failed CI is not relevant to this pr

PR #226, opened 2022-08-31
Add support for docking in lanes that have entry events

luca-della-vedova:
### Implemented feature

This PR addresses a TODO where having a `dock_name` property in a waypoint that had an entry or exit event would make the fleet adapter throw an exception.

It turns out there is no issue with exit events, but an implementation has been added for entry events by inserting a waypoint manually, basically the following:

```
A (entry event) -----> B
```

becomes

```
A (entry event) --> D (docking event) --> B
```

A and D overlap with each other. This has the effect of triggering the entry event first, then the docking event on the way to B.

This behavior is useful in cases where a system integrator wants to use docking in cases where there are entry events. A typical case is having a docking waypoint inside a lift to execute a special behavior to enter it.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/parse_graph.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/parse_graph.cpp
@@ -278,7 +278,7 @@ rmf_traffic::agv::Graph parse_graph(
         {
           // Add a waypoint and a lane leading to it for the dock maneuver
           // to be done after the entry event
-          const auto entry_wp = graph.get_waypoint(end);
+          const auto entry_wp = graph.get_waypoint(begin);
           auto& dock_wp = graph.add_waypoint(map_name, entry_wp.get_location());
 
           graph.add_lane(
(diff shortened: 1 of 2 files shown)


mxgrey:
@luca-della-vedova the change looks good to me, but I'm curious whether this has been tested on a graph that has a docking+event combo?

luca-della-vedova:
> @luca-della-vedova the change looks good to me, but I'm curious whether this has been tested on a graph that has a docking+event combo?

Yap! This is being tested with the TRL project to trigger docking when entering or exiting lifts, i.e. [here](https://github.com/TRL-RMF/trl/blob/9ffe88255ab43ad257cfa9528c9a22462e96bfa8/trl_demos_maps/maps/trl/trl.building.yaml#L1424) to enter and [here](https://github.com/TRL-RMF/trl/blob/9ffe88255ab43ad257cfa9528c9a22462e96bfa8/trl_demos_maps/maps/trl/trl.building.yaml#L1525) to exit. Without this PR it throws the exception, with the PR it triggers a docking when approaching the waypoints

mxgrey (approved):
Thanks for getting this to work! The exception was a dirty shortcut that I took a long time ago because I didn't have enough time to think about how best to do it, so I'm glad to see this situation becoming functional.

mxgrey (approved):

PR #228, opened 2022-09-08
Async behavior fixes

mxgrey:
This PR fixes many subtle issues related to async behaviors. Prior to this PR, there are many opportunities for race conditions to cause problematic outcomes, causing explosions in delay estimations, causing negotiations to hang permanently, or causing tasks to stop permanently with no recovery mechanism. There are also some memory consumption fixes in this PR, as there were a few ways in which the memory consumption of a fleet adapter would be able to grow indefinitely.

These issues should be fixed in this PR, confirmed by hours of intensive full-stack integration testing.

This PR depends on these other two:
* https://github.com/open-rmf/rmf_internal_msgs/pull/45
* https://github.com/open-rmf/rmf_traffic/pull/86

CI won't be able to pass until those are merged.


rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
```diff
@@ -227,6 +227,18 @@ class Dispatcher::Implementation
...
+    const std::string uri =
+      node->declare_parameter("server_uri", std::string());
+    if (!uri.empty())
+    {
+      RCLCPP_INFO(
+        node->get_logger(),
+        "API server URI: [%s]", uri.c_str());
+      server_uri = uri;
+    }
+
```
xiyuoh, line 241: Noticed a duplicate code here while trying to build the package

luca-della-vedova (approved):

PR #231, opened 2022-09-10
Improving CI

orensbruli:
Signed-off-by: Esteban Martinena <orensbruli@gmail.com>

### Implemented feature

Added execution matrix with ROS and Ubuntu versions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,76 +4,17 @@ on:
   pull_request:
   schedule:
     - cron: '23 0 * * *'
-defaults:
-  run:
-    shell: bash
+  workflow_dispatch:
+
 jobs:
   build_and_test:
-    strategy:
-      fail-fast: false
-      matrix:
-        ros_distribution:
-          - foxy
-          - galactic
-          - humble
-        ubuntu_distribution:
-          - focal
-          - jammy
-        exclude:
-          - ros_distribution: foxy
-            ubuntu_distribution: jammy
-          - ros_distribution: galactic
-            ubuntu_distribution: jammy
-          - ros_distribution: humble
-            ubuntu_distribution: focal
-
-    name: Build and test
-    runs-on: ubuntu-22.04
-    container:
-      image: rostooling/setup-ros-docker:ubuntu-${{ matrix.ubuntu_distribution }}-latest
-    steps:
-      - name: deps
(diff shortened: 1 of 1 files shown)


mxgrey:
It seems an upstream update to uncrustify has changed the behavior of the code style linter. I was going through the diff and fixing based on the recommendations, but I'm really floored by this one:

```
-    best_estimate = Info{cost, &setup};
+    best_estimate = Info{cost, &setup}
+  ;
```

It literally wants our code to look like

```
  best_estimate = Info{cost, &setup}
;
```

I've swallowed my pride and accepted many of uncrustify's whitespace requirements that I consider terrible styling, but I think I have to draw a line here. Since I plan on aggressively migrating this code base to Rust, I don't think it's worth anyone's time to figure out what uncrustify rules need to be changed to get rid of this bug, so I'll try to find some tags to tell uncrustify to ignore these code blocks.

Yadunund:
`ament_uncrustify` uses a vendored version of `uncrustify`. It seems that for the first time since `eloquent`, the vendored version has been bumped as seen in this PR https://github.com/ament/uncrustify_vendor/pull/27

If the vendored version has some serious bugs which have been patched, maybe we should submit a PR to `uncrustify_vendor` to include use the commit version with the fix.

orensbruli:
> It seems an upstream update to uncrustify has changed the behavior of the code style linter. I was going through the diff and fixing based on the recommendations, but I'm really floored by this one:
> 
> ```
> -    best_estimate = Info{cost, &setup};
> +    best_estimate = Info{cost, &setup}
> +  ;
> ```
> 
> It literally wants our code to look like
> 
> ```
>   best_estimate = Info{cost, &setup}
> ;
> ```
> 

Yeah I also found this. It looks like some kind of bug. I forced an old version installation to get over it. These were my personal notes:

> Failing test because of a weird return {}; https://github.com/open-rmf/rmf_task/runs/8285620339?check_suite_focus=true#step:5:1757
> Looks like it can be a minor problem in uncrustify.
> Also fount it can be a problem with new versions of uncrustify
> Downgrading version of uncrustify on jammy installations works.
> Issue https://github.com/uncrustify/uncrustify/issues/3191
> Workaround https://github.com/open-rmf/rmf_ci_templates/blob/main/.github/workflows/reusable_build.yaml#L44

mxgrey (approved):

PR #235, opened 2022-09-13
EasyFullControl

xiyuoh:
This PR continues from #189 to implement an `EasyFullControl` API for users to create their own fleet adapters more easily.

This PR can be tested with https://github.com/open-rmf/rmf_demos/pull/158. You can run any full control demo world with `easy_fleet:=1`:
``` bash
ros2 launch rmf_demos_gz_classic office.launch.xml easy_fleet:=1
```

Further to-do:
- [x] Transformer between RMF and robot coordinates
- [x] Update with fixes from https://github.com/open-rmf/rmf_demos/pull/157
- [x] Python bindings


xiyuoh:
Thanks for the suggestion! I understand better now, the callback relationship has been inverted in [8843f44](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8843f4416babcd91ff489686517bdaf7f194d67b).

I see that an `update_position()` method is added to ActionExecution to update the robot's map name and position, which is not necessary as we already have a GetStateCallback retrieving the robot's RobotState on every update. Custom actions are also using this callback to update current map and position.
Instead, the remaining parameters that need to be updated are request_replan and remaining_time, so I've modified the `update_position()` to an `update_request()` instead.

I am also curious, why shouldn't we use ActionExecution's existing `update_remaining_time()` method to update time estimates for navigation requests?

Yadunund:
@xiyuoh looks like CI is failing with this build error
```
  /__w/rmf_ros2/rmf_ros2/ros_ws/src/0ls7u2nzuzcr/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_RobotUpdateHandle.hpp:38:10: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 4 were provided
    struct Data
           ^
  3 warnings and 2 errors generated.
  gmake[2]: *** [CMakeFiles/rmf_fleet_adapter.dir/build.make:174: CMakeFiles/rmf_fleet_adapter.dir/src/rmf_fleet_adapter/agv/EasyFullControl.cpp.o] Error 1
  gmake[2]: *** Waiting for unfinished jobs....
  gmake[1]: *** [CMakeFiles/Makefile2:188: CMakeFiles/rmf_fleet_adapter.dir/all] Error 2
  gmake: *** [Makefile:146: all] Error 2
```

xiyuoh:
@Yadunund Oh I think the latest PR added a `worker` field to Data that wasn't implemented here. Is exposing the RobotContext (so that we can access the `worker` from `follow_new_path`) a good solution? The current approach is to create an ActionExecution object to be passed to NavigationRequest for progress tracking.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -130,7 +130,12 @@ class RobotUpdateHandle
   class ActionExecution
   {
   public:
-    /// Update the amount of time remaining for this action
+    /// Update navigation or docking replan requests and remaining time estimate.
+    void update_request(
```
Yadunund, line 134: I agree we don't need an `update_position()` given the `GetStateCallback` that is registered for the robot. But we also do not need this `update_request()` function. Users can call the `update_remaining_time()` that already exists below. 
To request for a replan, we can have a separate `bool replan();` function.
We could also add a function to enable/disable automatic interpolation of remaining duration which was the point of having the `std::optional<rmf_traffic::Duration>` iirc

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -147,12 +152,17 @@ class RobotUpdateHandle
...
+    /// Trigger this when the action is successfully finished.
+    /// No other functions in this ActionExecution instance will
+    /// be usable after this.
     void finished();
 
     /// Returns false if the Action has been killed or cancelled
     bool okay() const;
 
+    /// Get the RobotUpdateHandle
+    std::optional<std::shared_ptr<RobotUpdateHandle>> handle() const;
```
Yadunund, line 164: The original intention to retrieve the handle was for users to be able to update the schedule trajectory of the robot while it is performing an action. But I think we should not provide this capability as it could lead to unintended updating of states. Instead, we can provide an API
```
    /// Override the schedule to say that the robot will be following a certain
    /// trajectory. This should not be used while tasks with automatic schedule
    /// updating are running, or else the traffic schedule will have jumbled up
    /// information, which can be disruptive to the overall traffic management.
    void override_schedule(rmf_traffic::Trajectory trajectory);
```

The implementation of this function can call the Unstable::get_participant() API and update the itinerary with the trajectory provided.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
```diff
@@ -813,6 +813,14 @@ void RobotUpdateHandle::Unstable::set_lift_entry_watchdog(
   }
 }
 
+//==============================================================================
+void RobotUpdateHandle::ActionExecution::update_request(
```
Yadunund, line 817: To be updated as described above.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
```diff
@@ -0,0 +1,2175 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
```
Yadunund, line 2:
```suggestion
 * Copyright (C) 2023 Open Source Robotics Foundation
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
```diff
@@ -0,0 +1,2175 @@
...
+      RCLCPP_INFO(
+        node->get_logger(),
+        "Fleet [%s] is configured to perform Clean tasks",
+        config.fleet_name().c_str()
+      );
+    }
+  }
+
+  easy_adapter->_pimpl->fleet_handle->consider_composed_requests(consider_all);
+  for (const auto [action, consider] : config.action_consideration())
```
Yadunund, line 2003:
```suggestion
  for (const auto& [action, consider] : config.action_consideration())
```

Yadunund (changes requested):
Here's some feedback

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
xiyuoh: [8765d80](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8765d80367be1d29ab883f1cd863a615712cd214)

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
xiyuoh: [8765d80](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8765d80367be1d29ab883f1cd863a615712cd214)

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,445 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
```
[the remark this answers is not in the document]
xiyuoh: [8765d80](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8765d80367be1d29ab883f1cd863a615712cd214)

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,445 @@
...
+private:
+  EasyFullControl();
+  rmf_utils::unique_impl_ptr<Implementation> _pimpl;
+};
+
+using EasyFullControlPtr = std::shared_ptr<EasyFullControl>;
+
+/// A Transformation object that stores the transformation data needed to
+/// perform transformation between robot and RMF cartesian frames.
+struct Transformation
```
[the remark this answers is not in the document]
xiyuoh: Done in [8cda989](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8cda989031ad88656237913afa84ca23ac9eb16b), and moved the `transform` method into Transformation class in [7d2f03a](https://github.com/open-rmf/rmf_ros2/pull/235/commits/7d2f03aa26cdceefbef0111761cdd8f2cb6dd772)

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
xiyuoh: [8765d80](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8765d80367be1d29ab883f1cd863a615712cd214)

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
xiyuoh: Added a `void replan(bool request_replan)` in [8765d80](https://github.com/open-rmf/rmf_ros2/pull/235/commits/8765d80367be1d29ab883f1cd863a615712cd214). 
About the function to enable/disable remaining duration calculation, do you mean something like what is implemented [here](https://github.com/open-rmf/rmf_ros2/blob/feature/easy_full_control/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp#L757-L771)?

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -0,0 +1,445 @@
...
+    /// \param[in] nav_graph_path
+    ///   The path to a navigation path file that includes map information necessary
+    ///   to create a rmf_traffic::agv::Graph object
+    ///
+    /// \param[in] server_uri
+    ///   The URI for the websocket server that receives updates on tasks and
+    ///   states. If nullopt, data will not be published.
+    ///
+    /// \return A Configuration object with the essential config parameters loaded.
+    static std::shared_ptr<Configuration> make(
```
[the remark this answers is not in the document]
xiyuoh: Okay, updated and added a short line documenting this.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -152,6 +152,16 @@ class RobotUpdateHandle
...
 
+    /// Trigger this when you require a replan for your navigation or
+    /// docking request
+    void replan(bool request_replan);
+
+    /// Override the schedule to say that the robot will be following a certain
+    /// trajectory. This should not be used while tasks with automatic schedule
+    /// updating are running, or else the traffic schedule will have jumbled up
+    /// information, which can be disruptive to the overall traffic management.
+    void override_schedule(std::string map_name, rmf_traffic::Trajectory trajectory);
```
Yadunund: Sorry had a error in my previous comment
```
    void override_schedule(const std::string& map_name, const rmf_traffic::Trajectory& trajectory);
```

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -152,6 +152,16 @@ class RobotUpdateHandle
     /// (warning tier)
     void blocked(std::optional<std::string> text);
 
+    /// Trigger this when you require a replan for your navigation or
+    /// docking request
+    void replan(bool request_replan);
```
Yadunund, line 153: We don't have to pass in a bool arg for this function. Users can simply call this function to request for a replan. It can return a bool depending on if the replanning succeeded/failed.
`bool replan();`

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -147,7 +148,20 @@ class RobotUpdateHandle
...
-    /// Trigger this when the action is successfully finished
+    /// Trigger this when you require a replan for your navigation or
+    /// docking request
+    void replan(bool request_replan);
+
+    /// Override the schedule to say that the robot will be following a certain
+    /// trajectory. This should not be used while tasks with automatic schedule
+    /// updating are running, or else the traffic schedule will have jumbled up
+    /// information, which can be disruptive to the overall traffic management.
+    void override_schedule(std::string map_name,
```
Yadunund, line 159: Sorry had an error in my previous review comment.
```
    void override_schedule(const std::string& map_name, const rmf_traffic::Trajectory& trajectory);
```

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/Transformation.hpp
```diff
@@ -0,0 +1,73 @@
...
+  /// Get the scale of this Transformation
+  const double scale() const;
+
+  /// Get the translation of this Transformation
+  const Eigen::Vector2d& translation() const;
+
+  /// Helper function to transform between RMF and robot coordinate systems.
+  /// Depending on the Transformation defined, this function can be used to
+  /// transform robot coordinate system to RMF's coordinate system or vice versa.
+  const Eigen::Vector3d transform(
```
Yadunund, line 62: We should move this function outside this Transform class but in the same header/namespace and have it accept a transform object (Same as how it was in EasyFullControl.hpp).

Yadunund (changes requested):

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
xiyuoh: [afef64c](https://github.com/open-rmf/rmf_ros2/pull/235/commits/afef64c0caddacc9db50472b3e3f607d0c174065)

mxgrey:
> I see that an update_position() method is added to ActionExecution to update the robot's map name and position, which is not necessary as we already have a GetStateCallback retrieving the robot's RobotState on every update. 

Sorry for not making this part clear: I think we should get rid of `GetStateCallback` and `RobotState` entirely. I don't think it's a good idea for us to poll for this data when
1. Some of the fields won't change very often or ever (e.g. `name` and `charger_name`)
2. The fields inside of `RobotState` are largely independent, so it doesn't make sense to bundle them
3. The user is forced to manage mutexes in their implementation of the `GetStateCallback` to make sure they can collect and return all this information safely when reads/writes are happening across multiple threads

I'm going to experiment a bit with how to straighten out these details. Ideally I'd like us to have an API where the user (system integrator) never needs to touch a mutex, unlike the current situation where we've found that every function in the command handle of `rmf_demos_fleet_adapter` needs to be guarded by a mutex to avoid risk of serious race condition bugs.

Yadunund:
> > I see that an update_position() method is added to ActionExecution to update the robot's map name and position, which is not necessary as we already have a GetStateCallback retrieving the robot's RobotState on every update.
> 
> Sorry for not making this part clear: I think we should get rid of `GetStateCallback` and `RobotState` entirely. I don't think it's a good idea for us to poll for this data when
> 
> 1. Some of the fields won't change very often or ever (e.g. `name` and `charger_name`)
> 2. The fields inside of `RobotState` are largely independent, so it doesn't make sense to bundle them
> 3. The user is forced to manage mutexes in their implementation of the `GetStateCallback` to make sure they can collect and return all this information safely when reads/writes are happening across multiple threads
> 
> I'm going to experiment a bit with how to straighten out these details. Ideally I'd like us to have an API where the user (system integrator) never needs to touch a mutex, unlike the current situation where we've found that every function in the command handle of `rmf_demos_fleet_adapter` needs to be guarded by a mutex to avoid risk of serious race condition bugs.

That's a lot better 👍🏼

mxgrey (approved):

Yadunund (approved):

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -368,24 +444,7 @@ class RobotUpdateHandle
...
-    public:
-      /// Stop being stubborn
-      void release();
-
-      class Implementation;
-    private:
-      Stubbornness();
-      rmf_utils::impl_ptr<Implementation> _pimpl;
-    };
+    using Stubbornness = Stubbornness;
```
arjo129, line 447: I'm getting a compiler error from this:
```
In file included from /home/arjoc/ksoe_ws/src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.hpp:22,
                 from /home/arjoc/ksoe_ws/src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/Negotiator.hpp:23,
                 from /home/arjoc/ksoe_ws/src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/Negotiator.cpp:18:
/home/arjoc/ksoe_ws/src/rmf/rmf_ros2/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp:447:11: error: declaration of ‘using Stubbornness = class rmf_fleet_adapter::agv::RobotUpdateHandle::Stubbornness’ changes meaning of ‘Stubbornness’ [-fpermissive]
  447 |     using Stubbornness = Stubbornness;
      |           ^~~~~~~~~~~~
```

mxgrey (approved):

mxgrey (approved):

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
arjo129: LGTM

PR #243, opened 2022-10-07
Add find_package for vendored project.

nuclearsandwich:
The vendor package for nlohmann_json_schema_validator was previously exporting dependency info for that package, however that is not the recommended workflow for vendor packages which are ideally as transparent as possible.

This PR should be held until https://github.com/open-rmf/nlohmann_json_schema_validator_vendor/pull/10 is merged.


orensbruli (approved):

orensbruli:
@nuclearsandwich it looks like there's some problem with the sign of this commit?

orensbruli:
Merged here https://github.com/open-rmf/nlohmann_json_schema_validator_vendor/pull/12

PR #244, opened 2022-10-08
Refinements to EasyFullControl

Yadunund:
This PR updates the EasyFullControl implementation.

TODO: Describe changes, add documentation, update bindings.


rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -175,7 +175,7 @@ class EasyFullControl::Configuration
   ///   The URI for the websocket server that receives updates on tasks and
   ///   states. If nullopt, data will not be published.
   Configuration(
-    const std::string& fleet_name,
+    std::string fleet_name,
```
Yadunund, line 178: Why is this a dangling reference? It previously allowed strings to be passed as both l and r values?

PR #247, opened 2022-11-08
Update easy full control

xiyuoh:
This PR updates the `feature/easy_full_control` branch with the following:
- `Configuration::make` to easily create a Configuration object by parsing config and navigation graph file paths
- Modify `GoalCompletedCallback` to return the goal remaining time and replan request instead of updating the references
- Add `Transformation` struct and `transform` helper function


mxgrey:
I don't think a `std::unordered_map<std::string, bool>` is expressive enough for real use cases. Users of this API should be able to discern between which task requests the fleet adapter should accept based on the descriptions. This might not matter in our demos, but it will certainly matter for real-world use cases when things like inventory need to be accounted for when deciding which tasks to accept.

I've updated this PR to use `std::unordered_map<std::string, ConsiderRequest>` for task and action consideration settings. For convenience there's also a `consider_all()` function that allows users to easily get the original boolean behavior.

mxgrey (approved):

PR #253, opened 2022-12-08
Fix emergency response for waiting robots

mxgrey:
Previously when a robot is in responsive wait mode, it would not behave correctly when an emergency alarm is triggered.

This PR fixes the problem by cancelling the responsive wait behavior when an emergency alarm is activated.


luca-della-vedova (approved):
Works great, thanks!

PR #255, opened 2022-12-10
Fix gh_pages workflow

orensbruli:


orensbruli:
It should fix the gh_pages workflow, BUT I think it's not being used anymore (or anytime) and we could remove it.

PR #265, opened 2023-03-24
Fix priority assignment

Yadunund:
This PR fixes https://github.com/open-rmf/rmf_ros2/issues/266 by correcting the priority assignment when converting a json task request into an `rmf_task::Request`

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -235,7 +235,7 @@ std::shared_ptr<rmf_task::Request> FleetUpdateHandle::Implementation::convert(
   if (t_it != request_msg.end())
   {
     earliest_start_time =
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -251,15 +254,25 @@ std::shared_ptr<rmf_task::Request> FleetUpdateHandle::Implementation::convert(
         const auto& p_value = (*p_it)["value"];
         if (p_value.is_number_integer())
         {
+          // If we have an integer greater than 0, we assign a high priority.
+          // Else the priority will default to low.
           if (p_value.is_number_integer() && p_value.get<uint64_t>() > 0)
+          {
             priority = rmf_task::BinaryPriorityScheme::make_high_priority();
+          }
         }
```
mxgrey, line 263: I'll add an `else { invalid_schema = true; }` here to catch cases where the `"value"` isn't a number.

mxgrey (approved):
Thanks for fixing this so quickly!

PR #267, opened 2023-04-17
Adding initiator and request time to booking

aaronchongth:
### Implemented feature

Related to https://github.com/open-rmf/rmf-web/pull/702, and requires https://github.com/open-rmf/rmf_task/pull/81 and https://github.com/open-rmf/rmf_api_msgs/pull/35.

* To use new `rmf_task::Request` and `rmf_task::Task::Booking` constructors
* Propagated request time and requester information to `Dispatcher`, `TaskManager` and `FleetUpdateHandle`
* Added default argument to `set_task_planner_parameter` for python bindings of `FleetUpdateHandle`
* Updated full control fleet adapter's finishing task factories


rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -1234,18 +1234,28 @@ std::shared_ptr<Connections> make_fleet(
   const std::string finishing_request_string =
     node->declare_parameter("finishing_request", "nothing");
   rmf_task::ConstRequestFactoryPtr finishing_request = nullptr;
+  auto get_time =
+    []()
+    {
+      return std::chrono::steady_clock::now();
```
mxgrey, line 1240: We definitely do not want `steady_clock` here. That won't be a meaningful time stamp; it would just be the time since this node began running.

I would recommend using ROS2 time given by the ROS2 node's clock, or using `std::chrono::system_clock::now()` instead.

rmf_fleet_adapter_python/src/adapter.cpp
```diff
@@ -331,23 +332,38 @@ PYBIND11_MODULE(rmf_adapter, m) {
...
+      std::function<rmf_traffic::Time()> time_now = nullptr;
+      std::optional<std::string> planner_id = std::nullopt;
+      if (node)
+      {
+        time_now = [n = std::weak_ptr<rclcpp::Node>(node)]()
+        {
+          const auto node = n.lock();
+          if (!node)
+          {
+            return std::chrono::steady_clock::now();
```
mxgrey, line 347: If the ROS2 node isn't available, `system_clock` is more likely to give a time that makes sense, so I'd suggest using that instead.

mxgrey (changes requested):
I left some recommendations to not use `steady_clock` for timestamps, but once that's changed I'll be happy to approve.

rmf_fleet_adapter/src/full_control/main.cpp
aaronchongth: Good point, I've made the change to use node time when it's available, otherwise `system_clock`, https://github.com/open-rmf/rmf_ros2/pull/267/commits/6815ea70e6fd9ec4fa03c82523c2c134c1aa9f50

rmf_fleet_adapter_python/src/adapter.cpp
aaronchongth: gotcha, I did a tiny workaround to lump `system_clock` to `rmf_traffic::Time` which is based on `steady_clock`. Was there any specific reason we were using `steady_clock` for `rmf_traffic::Time` in the past instead of system?
https://github.com/open-rmf/rmf_ros2/pull/267/commits/6815ea70e6fd9ec4fa03c82523c2c134c1aa9f50

mxgrey (approved):

mxgrey (changes requested):
Hold on, I just noticed something that I'd like to tweak before we merge this.

mxgrey (approved):
I've removed the need to pass a node into `set_task_planner_params`. I believe https://github.com/open-rmf/rmf_demos/pull/180 should be unnecessary now.

aaronchongth:
Thanks for the changes! The node API is definitely the cleanest way to do it

PR #204, opened 2022-05-04
Fixes to solve pthread build errors.

decada-robotics:
Some edits to prevent build errors in the rmf_ros2 package.

Issue #148, opened 2021-12-20
linking error when building rmf_fleet_adapter


mayman99:
## Bug report

`
undefined reference to symbol 'pthread_rwlock_unlock@@GLIBC_2.2.5' make
` when building the package
### Required information:

- Operating system and version:
  - Ubuntu 20.04
- OpenRMF installation type:
  - source
- OpenRMF version or commit hash
  - latest, 66cd1dc
- ROS distribution and version:
  - 
- ROS installation type:
  - binary
- Package or library, if applicable:
  - rmf_fleet_adapter

### Description of the bug

fixed by adding `-pthread` to `experimental_lift_watchdog` target link libs
```
target_link_libraries(experimental_lift_watchdog
  PUBLIC
    -pthread
  PRIVATE
    ${rclcpp_LIBRARIES}
    ${rmf_fleet_msgs_LIBRARIES}
    ${std_msgs_LIBRARIES}
)
```

### Additional information

`rmf_fleet_adapter`

mxgrey:
Linkers are awfully mysterious.. I have no idea how you're the first one to be dealt this error.

There's a preferable fix to this problem than explicitly linking `libpthread`:

```
find_package(Threads REQUIRED)
...
target_link_libraries(target_name [PUBLIC|PRIVATE] Threads::Threads)
```

That should fix the linking problem while also being cross-platform.

Yadunund:
Potentially fixed by https://github.com/open-rmf/rmf_ros2/pull/204

PR #268, opened 2023-04-19
Explicitly link to the Threads target

mxgrey:
It seems the ROS2 build farm [is not happy](https://build.ros2.org/job/Rbin_ujv8_uJv8__rmf_fleet_adapter__ubuntu_jammy_arm64__binary/149/) with `rmf_fleet_adapter`, and it appears to have something to do with pthread in cmake. Usually this kind of issue is resolved by finding cmake's built in `Threads` package and linking to its `Threads::Threads` target. Hopefully this PR resolves the build farm issue.

### DIFF
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -13,6 +13,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(yaml-cpp REQUIRED)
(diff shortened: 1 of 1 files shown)


Yadunund (approved):
LGTM.

Although I'd prefer if we merged https://github.com/open-rmf/rmf_ros2/pull/204 in to give credit to "an earlier fix" but the author has not made the requested change, leaving the PR incomplete. 

But yeah this has been a [long standing issue](https://github.com/open-rmf/rmf_ros2/issues/148) so we should go ahead and merge this in.

PR #273, opened 2023-05-18
Put the action finished callback in a schedule instead of triggering immediately

mxgrey:
According to https://github.com/open-rmf/rmf/issues/349 there might be a data race related to `ActionExecution::finished`. This PR should reduce the surface area for where a data race may be occurring.


Yadunund (approved):

Yadunund (approved):
I moved the changelog entry for this PR into "Forthcoming" since 2.1.5 is already released. 
In the future, we can just update the changelogs when doing a version bump using [catkin_generate_changelogs](https://docs.ros.org/en/iron/How-To-Guides/Releasing/Subsequent-Releases.html#updating-changelog)

PR #283, opened 2023-06-15
WIP Add api for robot charging tasks

### LINKED ISSUE #261: Add API for direct robot charging requests
### Description

There is currently no API for system integrators to submit a robot charging request. This is especially problematic since direct task requests bypass the implicit charging task insertion logic, instead if the robot doesn't have enough battery to perform the direct task assignment RMF will print a warning but still send the robot.
For direct task requests, the system integrator is expected to keep track of whether a robot has enough battery and charge it if needed. However, since currently there is no API to do so, the only way to go at the moment is to set the `finishing_request` of the fleet adapter to `charge` then wait until the battery state of charge is high enough before submitting the task request.

### Implementation considerations

A schema for the charging task should be added [here](https://github.com/open-rmf/rmf_ros2/tree/main/rmf_fleet_adapter/schemas) and deserialization should be updated to parse it when adding standard tasks.

luca-della-vedova:
### Implemented feature

Closes #261.

This PR adds support for submitting robot charging requests through a `charge` task.
This task should only be used in direct task requests, since dispatching it to the robot that can charge the fastest doesn't really make a lot of sense.

It can be tested with [this rmf_demos branch](https://github.com/open-rmf/rmf_demos/compare/luca/charge_task) that adds a `dispatch_charge` task, with mandatory `fleet` and `robot` parameters to make sure it is a direct assignment.

### DIFF
--- a/rmf_fleet_adapter/schemas/event_description__charge.json
+++ b/rmf_fleet_adapter/schemas/event_description__charge.json
@@ -1,0 +1,9 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/open-rmf/rmf_ros2/main/rmf_fleet_adapter/schemas/event_description__charge.json",
+  "title": "Charge Event",
+  "description": "Charge a robot",
+  "type": "object",
+  "properties": {
+  }
+}
--- a/rmf_fleet_adapter/schemas/task_description__charge.json
+++ b/rmf_fleet_adapter/schemas/task_description__charge.json
@@ -1,0 +1,7 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/open-rmf/rmf_ros2/main/rmf_fleet_adapter/schemas/task_description__charge.json",
+  "title": "Charge Task",
+  "description": "Charge a robot",
+  "$ref": "event_description__charge.json"
+}
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -1104,6 +1104,7 @@ void FleetUpdateHandle::Implementation::add_standard_tasks()
     node->clock());
 
   tasks::add_charge_battery(
+    deserialization,
     *activation.task,
     activation.phase,
     *activation.event,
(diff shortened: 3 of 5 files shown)


rmf_fleet_adapter/schemas/event_description__charge.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/open-rmf/rmf_ros2/main/rmf_fleet_adapter/schemas/event_description__charge.json",
+  "title": "Charge Event",
```
mxgrey, line 4: I recommend we add "Battery" everywhere to make this more explicit, e.g. "Charge Battery Event", `event_description__charge_battery.json`, "Charge the battery of a robot".

If anyone develops a bull robot for a rodeo they might want "charge" to mean something else.

luca-della-vedova:
In https://github.com/open-rmf/rmf_ros2/pull/283/commits/d87b87bfdb952cc4cc581be5583d1cf6409cb1b4 an `unordered_set<std::string>` has been added to be checked when bidding for tasks to make sure fleet adapters don't bid on `charge_battery` tasks. Now `charge_battery` tasks can only be submitted through direct assignment and bidding will fail.
TODOs have been left in place to expose this as a parameter, for system integrators to choose which tasks they want to accept as direct but don't want to bid on.

rmf_fleet_adapter/schemas/event_description__charge_battery.json
```diff
@@ -0,0 +1,9 @@
+{
```
Yadunund, line 1: I don't think we necessarily need this `event_description` schema since we don't support any `ChargeBattery` events in `rmf_task`.
With this description, users can theoretically submit a composable task with a ChargeBattery event injected between other activities but the fleet adapter will not find the right builder from `rmf_task_sequence` to generate the `Model` and `Description` for this event.

Not a blocker imo but it would be good to test what happens when such a compose task is submitted to make sure any errors are gracefully handled.

rmf_fleet_adapter/schemas/event_description__charge_battery.json
```diff
@@ -0,0 +1,9 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/open-rmf/rmf_ros2/main/rmf_fleet_adapter/schemas/event_description__charge_battery.json",
+  "title": "Charge Battery Event",
+  "description": "Charge a robot's battery",
+  "type": "object",
+  "properties": {
```
Yadunund, line 7: I think right now the battery level to which the robot charges up to is statically defined in the fleet config. In the future it would be good to specify this % in the `properties` here and have the `TaskPlanner` accept this value in the `Options` it receives before it plans. Not a blocker.

PR #289, opened 2023-07-25
Configurable merge distances

xiyuoh:
This PR supplements #286 to add a few more configurable parameters to the fleet adapter:
- `max_merge_waypoint_distance`
- `max_merge_lane_distance`
- `min_lane_length`

As suggested, a default fleet value for each parameter can be configured in the fleet's config.yaml ([example](https://github.com/open-rmf/rmf_demos/blob/xiyu/configurable_merge_distance/rmf_demos/config/office/tinyRobot_config.yaml#L35-L37)), and robot-specific distances can be set as well ([example](https://github.com/open-rmf/rmf_demos/blob/xiyu/configurable_merge_distance/rmf_demos/config/office/tinyRobot_config.yaml#L42-L44)). To modify these values live, the `EasyRobotUpdateHandle` API is updated to accommodate getting and setting them.

These values are currently set independently since it might be more preferable to retrieve them individually, but I am wondering if it is better to take the params as optional values into a single function instead similar to `RobotContext`, such as
```cpp
void EasyFullControl::EasyRobotUpdateHandle::set_nav_params(
  std::optional<double> max_merge_waypoint_distance,
  std::optional<double> max_merge_lane_distance,
  std::optional<double> min_lane_length)
{
  // modify value if has_value()
  // ...
  context->set_nav_params(new_nav_params);
}
```

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
@@ -118,15 +118,24 @@ class EasyFullControl::RobotConfiguration::Implementation
 public:
   std::vector<std::string> compatible_chargers;
   std::optional<bool> responsive_wait;
+  std::optional<double> max_merge_waypoint_distance;
+  std::optional<double> max_merge_lane_distance;
+  std::optional<double> min_lane_length;
 };
 
 //==============================================================================
 EasyFullControl::RobotConfiguration::RobotConfiguration(
   std::vector<std::string> compatible_chargers,
-  std::optional<bool> responsive_wait)
+  std::optional<bool> responsive_wait,
+  std::optional<double> max_merge_waypoint_distance,
+  std::optional<double> max_merge_lane_distance,
+  std::optional<double> min_lane_length)
 : _pimpl(rmf_utils::make_impl<Implementation>(Implementation{
     std::move(compatible_chargers),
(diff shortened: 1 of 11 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
```diff
@@ -1400,6 +1461,101 @@ void EasyFullControl::EasyRobotUpdateHandle::update(
...
+}
+
+//==============================================================================
+void EasyFullControl::EasyRobotUpdateHandle::set_max_merge_waypoint_distance(
+  double distance)
+{
+  auto updater = _pimpl->updater;
+  if (!updater->handle)
+  {
+    return;
```
mxgrey, line 1487: It's not okay for these setters to silently have no effect. Users would be very confused if calling a function seems to randomly have no effect.

I'm going to refresh myself on what situations `updater->handle` might be a `nullptr` and see if we can tighten this up.

mxgrey: Fixed by [91a5270](https://github.com/open-rmf/rmf_ros2/pull/289/commits/91a5270b4da219b20ad9b95cf93e024f0a21a857)

mxgrey (approved):
Thanks for adding these parameters! We'll merge this in as soon as CI passes.

PR #301, opened 2023-09-15
Cancel automatic pending tasks that are removed during new assignments

aaronchongth:
### Fixed bug

Not exactly a bug, but at the moment pending automatic tasks are just removed from the queue without cancellation.
The original automatic task will remain as `queued` forever on the end of the `api-server`.

This publishes the cancellation of those pending tasks that will be removed when new assignments happen.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -925,7 +925,7 @@ void TaskManager::set_queue(
     // assignments.
     std::vector<std::string> new_automatic_task_ids;
     const std::vector<std::string> cancellation_labels =
-      {"New task assignments received."};
+    {"New task assignments received."};
     for (const auto& a : assignments)
     {
       if (a.request()->booking()->automatic())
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -920,6 +920,31 @@ void TaskManager::set_queue(
     {
       return;
     }
+
+    // Cancel pending automatic tasks that will be removed from the new
+    // assignments.
+    std::vector<std::string> new_automatic_task_ids;
```
Yadunund, line 926: We can get O(1) lookup in the second for loop below if we switch this container to an `unordered_set`.

```suggestion
    std::unordered_set<std::string> new_automatic_task_ids;
```

aaronchongth: thanks for catching that, https://github.com/open-rmf/rmf_ros2/pull/301/commits/9a6735de09a12e9bcb40d1017652162927b69a8c

Yadunund (approved):

PR #306, opened 2023-10-18
Change QoS for adapter_lift_requests and lift_requests to transient l…

xiyuoh:
This PR updates the QoS profile of LiftRequest messages for both `/adapter_lift_requests` and `/lift_requests`. The goal for this is to ensure that lift end session requests are reliably published and received by any lift adapters waiting to release control once the robot is done exiting the lift.


mxgrey (approved):
I've bumped the `keep_last` up to 100 since 10 might not be enough if there are too many lifts in one facility. I think 100 should be enough for any realistic facility, but this stands out as something that should be configurable in our next gen work.

PR #308, opened 2023-11-21
Update `GoToPlace` to allow finding nearest spot

arjo129:
### Implemented feature

This commit adds support for selecting the nearest spot on the same
floor. This behaviour is convenient when looking at things from a cancellation perspective, one may want to go to the nearest parking spot for instance.

This commit depends on https://github.com/open-rmf/rmf_task/pull/101

Note: While the schema supports contraints, we don't yet have an API to express such constraints, hence by default it will always use the same floor constraint.


xiyuoh:
Thanks for working on this! Tried it out with the demos script and it's working for me :+1: I see that having the final selected destination to be on the same map is a requirement in the code itself because of the lack of API for constraints. Right now it seems like if all the locations provided are on a different map from the robot, the first place provided will be the destination. Should we not make this behaviour a default and allow the robot to take lifts? Or is this only temporary until the constraints API are added in?

rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Patrol.cpp
```diff
@@ -67,8 +69,20 @@ void add_patrol(
             std::make_move_iterator(place.errors.begin()),
             std::make_move_iterator(place.errors.end()));
         }
-        auto desc = GoToPlace::Description::make_with_multiple(goals);
+        auto desc = GoToPlace::Description::make_for_one_of(goals);
         return {desc, errors};
```
xiyuoh, line 73:
```suggestion
```

xiyuoh: Move this line below, otherwise the description would return before accounting for the constraints

rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Patrol.cpp
```diff
@@ -67,8 +69,20 @@ void add_patrol(
...
+        if (constraints != msg.end())
+        {
+          for (const auto& constraint : constraints.value())
+          {
+            if (constraint["type"].get<std::string>() == "prefer_same_map")
+            {
+              desc->prefer_same_map(true);
+            }
+          }
+        }
```
xiyuoh, line 85:
```suggestion
        }
        return {desc, errors};
```

rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Patrol.cpp
```diff
@@ -67,8 +69,20 @@ void add_patrol(
...
-        auto desc = GoToPlace::Description::make_with_multiple(goals);
+        auto desc = GoToPlace::Description::make_for_one_of(goals);
         return {desc, errors};
+
+        const auto constraints = msg.find("constraints");
+        if (constraints != msg.end())
+        {
+          for (const auto& constraint : constraints.value())
+          {
+            if (constraint["type"].get<std::string>() == "prefer_same_map")
```
xiyuoh, line 80: Constraint property should be `category` instead of `type` according to the schema

xiyuoh (changes requested):

rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Patrol.cpp
mxgrey: I can't believe I messed up my own schema.

mxgrey:
Thanks for catching those constraint parsing errors @xiyuoh . I've made the fixes and run simulation tests to confirm that the `prefer_same_map` option is working as intended.

mxgrey:
I've opened a related PR for `rmf_demos_tasks` to support the new `one_of` option: https://github.com/open-rmf/rmf_demos/pull/200

arjo129:
On another note for code readability's sake, should we move the GoToPlace description out of `patrol.cpp` and into `GoToPlace.cpp`. It would greatly increase code readability for contributors.

mxgrey:
> should we move the GoToPlace description out of patrol.cpp and into GoToPlace.cpp.

To keep the diff down and get this merged quickly let's keep it as-is, but I'm open to shuffling that around if we ever touch this code again.

PR #309, opened 2023-11-26
Apply inverse transforms in Transformation class

xiyuoh:
This PR updates the transformation implementation in EasyFullControl between robot and RMF coordinates, and also fixes a bug in the `apply_inverse` function inside the Transformation class.

These updates are required for users relying on the transformation implementation in EFC, especially those using RMF binaries.


mxgrey:
Closing since these changes are accounted for in #310 which has been merged.

PR #316, opened 2024-01-08
Issue relocalization command if the robot is on the wrong map

mxgrey:
There is a rare edge case when a replan happens while a robot is moving between floors: the new commanded path will be based on a floor that the robot is not yet localized to.

With this PR, the EasyFullControl adapter will check if the new path is on the same map that the robot most recently reported being on. If there's a difference then it will send a localization command first.


xiyuoh (dismissed):
Tested this PR out over some time with different scenarios, map switches have been triggered correctly as intended.

xiyuoh (approved):

PR #323, opened 2024-01-31
Explicitly specify all qos depth

koonpeng:
## Bug fix

Thanks to @luca-della-vedova for pointing out that using the system default qos might not be appropriate as that depends on the RMW implementation, particularly, for cyclone, it defaults to history depth 1, making it very likely for messages to be silently discarded.

This series of PRs explicitly set a history depth of 10 to all topics and services that does not already set the history depth, those that already set the history depth are untouched. In total 69 instances across rmf codebases are modified, I tested this briefly, so far I haven't encountered any problems yet.

@aaronchongth This might fix some of the "unexplained" issues.

For reference, this is the regex used

search: `(rclcpp::SystemDefaultsQoS\(\)(?![^;,]*keep_last\())((?:.|\n)*?[;,])`
replace: `$1.keep_last(10)$2`

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
@@ -34,7 +34,7 @@ std::shared_ptr<Node> Node::make(
   auto node = std::shared_ptr<Node>(
     new Node(std::move(worker), node_name, options));
 
-  auto default_qos = rclcpp::SystemDefaultsQoS();
+  auto default_qos = rclcpp::SystemDefaultsQoS().keep_last(10);
   default_qos.keep_last(100);
   auto transient_qos = rclcpp::SystemDefaultsQoS()
     .reliable().keep_last(100).transient_local();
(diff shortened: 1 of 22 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
```diff
@@ -34,7 +34,7 @@ std::shared_ptr<Node> Node::make(
   auto node = std::shared_ptr<Node>(
     new Node(std::move(worker), node_name, options));
 
-  auto default_qos = rclcpp::SystemDefaultsQoS();
+  auto default_qos = rclcpp::SystemDefaultsQoS().keep_last(10);
   default_qos.keep_last(100);
```
luca-della-vedova, line 38: This should be reverted (maybe put in a single line to make more readable?

aaronchongth (commented):
Thanks for looking into this! Just tried it and multiple dispatches at the same time is working. Only the review comment by @luca-della-vedova pending

aaronchongth (approved):
LGTM, thanks!

PR #325, opened 2024-02-05
Adds a simple parking spot management system. 

arjo129:
### Implemented feature

This commit adds a stupidly simple parking spot system. It ensures the
next location a robot goes to is not occupied. It also publishes
currently free parking spots. When combined with https://github.com/open-rmf/rmf_ros2/pull/308 it allows a robot
to move to the nearest free spot instead of the nearest spot only.

note: This PR still needs more testing

Depends on:
- https://github.com/open-rmf/rmf_internal_msgs/pull/64

Known issues:
- When the robots start in simulation not all robots seem to be running
  the idle task, hence some robot dont end up claiming parking spots.


luca-della-vedova:
I got it to segfault in a fairly reproducible way, I did the following on office world:

```
ros2 run rmf_demos_tasks dispatch_go_to_place -p lounge --use_sim_time
# Wait until the robot is on the way back to its designated charger
ros2 run rmf_demos_tasks dispatch_go_to_place -p tinyRobot2_charger --use_sim_time
```

This is the log

```
[fleet_adapter-17] [INFO] [1715866344.935831555] [tinyRobot_command_handle]: Commanding [tinyRobot2] to navigate to [19.89569926 -3.40715006  0.39677206] on map [L1]: cmd_id 2
[rviz2-10] 
[rviz2-10] ** (rviz2:2893259): WARNING **: 21:32:28.508: atk-bridge: get_device_events_reply: unknown signature
[fleet_adapter-17] [INFO] [1715866349.558899088] [tinyRobot_fleet_adapter]: Beginning a new parking task for robot [tinyRobot/tinyRobot2]
[fleet_adapter-17] [INFO] [1715866349.559667818] [tinyRobot_fleet_adapter]: Selecting a new go_to_place location from [1] choices for robot [tinyRobot/tinyRobot2]
[fleet_adapter-17] [INFO] [1715866349.559716868] [tinyRobot_fleet_adapter]: Got distance from [28] as 6.920327
[fleet_adapter-17] [ERROR] [1715866349.559834748] [tinyRobot_fleet_adapter]: Unable to find a path to any of the goal options [[place:tinyRobot2_charger]] for [tinyRobot/tinyRobot2]
[fleet_adapter-17] [ERROR] [1715866349.560428308] [tinyRobot_fleet_adapter]: Got ticket issueing claim
[fleet_adapter-17] [ERROR] [1715866349.560583908] [tinyRobot_fleet_adapter]: Claim issued
[queue_manager-14] [INFO] [1715866349.560666328] [rmf_chope_node]: Allocating 28 to 4
[fleet_adapter-17] [INFO] [1715866349.561145268] [tinyRobot_fleet_adapter]: Selecting a new go_to_place location from [1] choices for robot [tinyRobot/tinyRobot2]
[fleet_adapter-17] [INFO] [1715866349.561259908] [tinyRobot_fleet_adapter]: Planning for [tinyRobot/tinyRobot2] to [tinyRobot2_charger] from one of these locations:
[fleet_adapter-17]  -- lane 24: { L1 <  18.729 -3.89598> [patrol_A2] } -> { L1 < 19.8957 -3.40715> [lounge] } | location < 19.8835 -3.41239> | orientation 22.7334
[fleet_adapter-17]  -- lane 25: { L1 < 19.8957 -3.40715> [lounge] } -> { L1 <  18.729 -3.89598> [patrol_A2] } | location < 19.8835 -3.41239> | orientation 22.7334
[fleet_adapter-17] [INFO] [1715866349.564905018] [tinyRobot_fleet_adapter]: Executing go_to_place [tinyRobot2_charger] for robot [tinyRobot/tinyRobot2]
[fleet_adapter-17] [ERROR] [1715866349.566171208] [tinyRobot_fleet_adapter]: Releasing waypoint
[queue_manager-14] [INFO] [1715866349.566286968] [rmf_chope_node]: Releasing ticket for 3
[fleet_adapter-17] [INFO] [1715866349.567795628] [tinyRobot_command_handle]: Commanding [tinyRobot2] to navigate to [18.72901862 -3.8959816  -0.6960909 ] on map [L1]: cmd_id 3
[rmf_task_dispatcher-13] [INFO] [1715866352.777068445] [rmf_dispatcher_node]: Add Task [compose.dispatch-1] to a bidding queue
[rmf_task_dispatcher-13] [INFO] [1715866352.913008243] [rmf_dispatcher_node]:  - Start new bidding task: compose.dispatch-1
[fleet_adapter-17] [INFO] [1715866352.913647203] [tinyRobot_fleet_adapter]: [Bidder] Received Bidding notice for task_id [compose.dispatch-1]
[fleet_adapter-17] [INFO] [1715866352.914766214] [tinyRobot_fleet_adapter]: Planning for [2] robot(s) and [1] request(s)
[fleet_adapter-17] [INFO] [1715866352.916380434] [tinyRobot_fleet_adapter]: Submitted BidProposal to accommodate task [compose.dispatch-1] by robot [tinyRobot2] with new cost [34.054840]
[rmf_task_dispatcher-13] [INFO] [1715866355.113066131] [rmf_dispatcher_node]: Determined winning Fleet Adapter: [tinyRobot], from 1 responses
[rmf_task_dispatcher-13] [INFO] [1715866355.113148461] [rmf_dispatcher_node]: Dispatcher Bidding Result: task [compose.dispatch-1] is awarded to fleet adapter [tinyRobot], with expected robot [tinyRobot2].
[fleet_adapter-17] [INFO] [1715866355.113777561] [tinyRobot_fleet_adapter]: Bid for task_id [compose.dispatch-1] awarded to fleet [tinyRobot]. Processing request...
[fleet_adapter-17] [INFO] [1715866355.114370381] [tinyRobot_fleet_adapter]: Canceling go_to_place for robot [tinyRobot/tinyRobot2]
[ign-18] [INFO] [1715866355.125845791] [slotcar_tinyRobot2]: [tinyRobot2] already received request [3] -- continuing as normal
[ERROR] [fleet_adapter-17]: process has died [pid 2893273, exit code -11 [...]
```

I wonder if the "try to reserve a resource that is already reserved" is a corner case that causes this instability?

arjo129:
Fixed in [18d2f3e](https://github.com/open-rmf/rmf_ros2/pull/325/commits/18d2f3e58b76a25398ebdf478a2dd72ee4a3944c). It was an error in the cancellation behaviour.

arjo129:
Oh shoot. Messed up history. Will open new PR.

arjo129:
I broke something again... :facepalm:

arjo129:
This is ready for review. I've added a readme so its easier to test. Contents are here:


"Chope" is the very singaporean act of throwing an item on a table and marking it at yours before going to get your food. The goal of this package is to provide the same functionality in RMF.
Before a robot goes to its next destination the fleet adapter asks the chope-node if its next destination is available. If the destination is available the robot will immediately proceed to the next destination, otherwise it will allocate a free parking spot for the robot to wait at till the next destination becomes available in a FIFO manner. If you need more advanced methods than FIFO, feel free to hack this package. To enable the use of this package you need to add the following to your fleet configuration yaml at the fleet level and run the chope node:
```yaml
  use_parking_reservations: True
```
We recommend disabling `responsive_wait` when you do so. An example is available in `rmf_demos`.

Some immediate limitations of the chope node is that it is limited to scenarios with a single navigation graph. Overcoming this limitation should not be too much work now that the basic infrastructure is in place.

## Expected Behaviour

### Basic queueing
You should never need to interact with this package directly. Rather, you can simply dispatch a task using the traditional methods. If you have `rmf_demos` installed you can run the
office world. We can start by commanding the `tinyRobot2` to go to the pantry.
```bash
ros2 run rmf_demos_tasks dispatch_go_to_place -p pantry -F tinyRobot2 --use-sim-time
```
The robot should proceed as expected. We can then ask `tinyRobot1` to also go to the pantry. Nothing should happen as the pantry is already occupied and `tinyRobot1` is at its parking spot.
```bash
ros2 run rmf_demos_tasks dispatch_go_to_place -p pantry -F tinyRobot1 --use-sim-time
```
We can ask `tinyRobot2` to move to the lounge after this.
```bash
ros2 run rmf_demos_tasks dispatch_go_to_place -p lounge -F tinyRobot2 --use-sim-time
```
This should immediately trigger 2 things:
1. `tinyRobot2` will move to the lounge
2. `tinyRobot1` will proceed to move to the pantry taking tinyRobot2's place.

### Moving to a waitpoint
If we continue from the previous example. Lets ask `tinyRobot1` and `tinyRobot2` to swap places starting at the pantry and lounge respectively.
```bash
ros2 run rmf_demos_tasks dispatch_go_to_place -p pantry -F tinyRobot1 --use-sim-time
```
You should see tinyRobot1 move to`tinyRobot2_charger` as `lounge` is not a designated wait point. You can then command `tinyRobot2` to the lounge.
```bash
ros2 run rmf_demos_tasks dispatch_go_to_place -p lounge -F tinyRobot2 --use-sim-time
```
`tinyRobot1` and `tinyRobot2` should now proceed to swap places.

If you have the same number or more waitpoints than robots you should never have a deadlock.

## Protocol Used Behind the Scenes

The chope node has no idea about the state of the graph it simply maintains a list of available parking spots internally. The fleet adapter does most of the heavy lifting in the `GoToPlace`. When a `GoToPlace` event is started we first check if the robot is already at one of the target locations. If we are at the target location, no comms is needed. Otherwise the fleetadapter submits a `ReservationRequest` message with the desired end parameters.  The parameters are listed in terms of cost (where the cost function is distance). The chope node will issue a ticket for said request.When ready to proceed, send a claim message witht ht e ticket and with the list of potential waiting points ordered by distance. The chope node will then try to award the lowest cost target location. If it can't it will award the lowest cost waiting point to the robot. The fleet adapter will release the previous location as it starts to move to its next location.

## Known Issues
1. We can only use one nav graoh with the chope node (it should be easy to resolve this with a mapping between nav graphs but for now its not part of the initial release).
2. At start up if there is no idle task, the chope node will not know where the robots are. It is adviced to send 1 `GoToPlace` task for every robot that is added to the world.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
```diff
@@ -422,60 +487,71 @@ std::optional<rmf_traffic::agv::Plan::Goal> GoToPlace::Active::_choose_goal(
...
-  auto lowest_cost = std::numeric_limits<double>::infinity();
-  std::optional<std::size_t> selected_idx;
-  for (std::size_t i = 0; i < _description.one_of().size(); ++i)
+
+  if (_context->_parking_spot_manager_enabled())
   {
-    const auto wp_idx = _description.one_of()[i].waypoint();
-    if (only_same_map)
+    RCLCPP_INFO(_context->node()->get_logger(),
+      "Requesting Chope Node For Time To Progress");
```
luca-della-vedova, line 494: So if the parking spot manager is enabled this function effectively does nothing and always returns a `std::nullopt`?

rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.hpp
```diff
@@ -103,13 +106,22 @@ class GoToPlace : public rmf_task_sequence::Event
     void kill() final;
 
   private:
+    enum class ReservationState
+    {
+      Pending=0,
+      Requested=1,
+      RecievedResponse=2
```
luca-della-vedova, line 113: Typo

rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.hpp
```diff
@@ -103,13 +106,22 @@ class GoToPlace : public rmf_task_sequence::Event
...
+    ReservationState _current_reservation_state = ReservationState::Pending;
+
 
     Active(Description description);
 
     void _schedule_retry();
 
     std::optional<rmf_traffic::agv::Plan::Goal> _choose_goal(
-      bool only_same_map) const;
+      bool only_same_map);
```
luca-della-vedova, line 124: I haven't looked in depth but it seems this function should still be able to be `const`? It seems to just iterate through possible goals and choose the closest one

rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
```diff
@@ -402,7 +467,7 @@ std::string wp_name(const agv::RobotContext& context)
 
 //==============================================================================
 std::optional<rmf_traffic::agv::Plan::Goal> GoToPlace::Active::_choose_goal(
-  bool only_same_map) const
+  bool only_same_map)
 {
   auto current_location = _context->location();
   auto graph = _context->navigation_graph();
```
luca-della-vedova, line 473: It should have been part of the original PR but I believe this would also create a copy of the whole robot's navgraph, which can be quite significant. We should probably do `const auto&` variables instead also to make sure we don't edit them by accident

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ChopeNegotiator.hpp
```diff
@@ -0,0 +1,350 @@
...
+        // Order wait points by the distance from the destination.
+        std::vector<std::tuple<double, std::string, rmf_traffic::agv::Plan::Goal>>
+          waitpoints_order;
+        for (std::size_t wp_idx = 0;
+        wp_idx < self->_context->navigation_graph().num_waypoints(); wp_idx++)
+        {
+          const auto wp = self->_context->navigation_graph().get_waypoint(
+            wp_idx);
+
+          auto name = wp.name();
```
luca-della-vedova, line 90: Same as other places, here we would be pretty much copying the whole navgraph, should capture by const reference whenever possible

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ChopeNegotiator.hpp
```diff
@@ -0,0 +1,350 @@
...
+    RCLCPP_INFO(context->node()->get_logger(),
+      "Sending chope request");
+    negotiator->make_request(same_map);
+    return negotiator;
+  }
+
+ChopeNodeNegotiator()
+  {
+
+  }
```
luca-della-vedova, line 213: It seems in this case having a default constructor that does nothing might be a bit dangerous, we are using the `make` function to correctly initialize a negotiator and offering this option seems more of a footgun than anything else

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ChopeNegotiator.hpp
```diff
@@ -0,0 +1,350 @@
...
+private:
+  enum class ReservationState
+  {
+    Pending=0,
+    Requested=1,
+    RecievedResponceProceedWaitPoint=2,
+    RecievedResponseProceedImmediate=3
```
luca-della-vedova, line 220: Typo (I'd suggest just doing a search and replace in the whole diff for `Recieved -> Received`).

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ChopeNegotiator.hpp
```diff
@@ -0,0 +1,350 @@
...
+            _context->node()->get_logger(),
+            "Got distance from [%lu] as %f",
+            wp_idx,
+            result->cost());
+
+          if (result->cost() < lowest_cost)
+          {
+            selected_idx = i;
+            lowest_cost = result->cost();
+          }
```
luca-della-vedova, line 292: I noticed that before this PR there was only one call of `quickest_path` in the whole `rmf_ros2` codebase, while now there are 4 and all the new ones seem to do one flavor of "iterates through waypoints, get quickest path, select minimum or sort by distance), it is quite a fair bit of duplicated code

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ChopeNegotiator.hpp
```diff
@@ -0,0 +1,350 @@
...
+    RecievedResponceProceedWaitPoint=2,
+    RecievedResponseProceedImmediate=3
+  };
+
+
+
+  void make_request(bool only_same_map)
+  {
+    auto current_location = _context->location();
+    auto graph = _context->navigation_graph();
```
luca-della-vedova, line 228: Same comment for const reference as this would copy the whole navgraph

luca-della-vedova (changes requested):
I tried to look at the code and left some suggestions.
I feel there is a fair bit of cleanup / refactoring that could be done to make the implementation cleaner, the main blocker I would hold is on creating a new public API that is not guarded by any sort of `unstable` flag for enabling a prototype system.

I didn't leave nits inline to avoid creating 100 messages but it would be great if throughout the PR the following were addressed:

* Removing irrelevant formatting changes (i.e. some newlines here and there, retabbing).
* Making sure the newly included headers are coherently ordered (i.e. all the C headers together in alphabetical order, rather than mixed with RMF headers).
* Making sure there are no spelling mistakes in variable names / printouts.
* Making sure we only do `auto variable` when we want to actually create an editable copy of a variable, and do `const auto` for non editable copies (or trivially copyable types such as integers) and `const auto&` for any non editable reference to a complex type, such as waypoints, navgraphs.
* Avoiding too much code duplication (I left some comments inline but there seem to be at least 4 very similar implementations of "go through some waypoints, find the closest one/s"

rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
arjo129: yep!

rmf_fleet_adapter/src/rmf_fleet_adapter/events/EmergencyPullover.cpp
```diff
@@ -22,10 +22,44 @@
...
+std::string wp_name(
+  const agv::RobotContext& context,
+  const rmf_traffic::agv::Plan::Goal& goal)
+{
+  const auto& g = context.planner()->get_configuration().graph();
+  const auto& wp = g.get_waypoint(goal.waypoint());
+  if (wp.name())
+    return *wp.name();
+
+  return "#" + std::to_string(goal.waypoint());
```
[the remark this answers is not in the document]
arjo129: Good catch.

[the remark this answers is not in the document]
arjo129: Nevermind, these functions are used for logging. They dont affect the reservation system.

arjo129:
I'm able to reproduce. the bug. I know why it's happening. Gimme a few hours to fix it.

arjo129:
OK, the deadlock that you were facing should be resolved now. You can give it another trial. You'll need to pull the demos again cause of some obstructions in the office world.

xiyuoh (commented):
Tested with the fix, after making the build work with `msg.ticket.ticket_id` (see below) the `supplies` waypoint was freed up as it should! However, when the robots are swapping positions (tinyRobot1 to `tinyRobot1_charger` and tinyRobot2 to `supplies`), they get stuck in this back and forth motion shown in the recording with many replanning printouts. I suspect it has something to do with cost? It went on for a few minutes before they eventually resolve the negotiation.

https://github.com/user-attachments/assets/ba0867a0-2065-4241-94fa-3d67a20e5cda

I managed to break it again; after they resolved this back and forth movement and eventually swapped places (tinyRobot1 now at `tinyRobot1_charger`, tinyRobot2 at `supplies`), I submitted the following patrol tasks one after another:
- tinyRobot2 to `tinyRobot1_charger`
- tinyRobot1 to `supplies`

tinyRobot2 goes to `tinyRobot2_charger` as a wait point, and tinyRobot1 tries to head towards `supplies`. But it got lost right outside the `coe` door and both robots get stuck in place. Logs available [here](https://gist.github.com/xiyuoh/202af530ae469088ad760b71ba40cc60) from when the direct patrol tasks were sent.

Apart from those, I took a look at the code. I might be missing something here, a little curious to understand why a `ReservationRequest` is issued and requires a `ReservationResponse` before the fleet adapter makes a `ReservationClaim`? Looks like the only thing the chope node's `on_request` does is to assign a ticket ID and store the request, is it possible to combine `ReservationRequest` and `ReservationClaim`, i.e. assign a ticket ID upon a new claim request, and do away with the response?

rmf_chope_node/src/main.cpp
```diff
@@ -0,0 +1,568 @@
...
+  rmf_chope_msgs::msg::Ticket get_new_ticket(
+    const rmf_chope_msgs::msg::RequestHeader& request_header)
+  {
+    rmf_chope_msgs::msg::Ticket ticket;
+    ticket.header = request_header;
+    ticket.ticket_id = _last_issued_ticket_id;
+    RobotIdentifier robot_identifier {
+      request_header.robot_name,
+      request_header.fleet_name
+    };
```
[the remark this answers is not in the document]
arjo129: More old stuff.

arjo129:
> Tested with the fix, after making the build work with `msg.ticket.ticket_id` (see below) the supplies waypoint was freed up as it should! However, when the robots are swapping positions (tinyRobot1 to `tinyRobot1_charger` and tinyRobot2 to supplies), they get stuck in this back and forth motion shown in the recording with many replanning printouts. I suspect it has something to do with cost? It went on for a few minutes before they eventually resolve the negotiation.

This seems to be planner problem (I probably should merge the latest main branch in). The reservation system is working as expected.

> `tinyRobot2` goes to tinyRobot2_charger as a wait point, and tinyRobot1 tries to head towards supplies. But it got lost right outside the coe door and both robots get stuck in place. Logs available [here](https://gist.github.com/xiyuoh/202af530ae469088ad760b71ba40cc60) from when the direct patrol tasks were sent.

I'm not sure why the robots got lost. I'm also not sure what the best behaviour when lost is supposed to be. It seems that the `tinyRobot1` is allocated to `supplies` and 

> Apart from those, I took a look at the code. I might be missing something here, a little curious to understand why a ReservationRequest is issued and requires a ReservationResponse before the fleet adapter makes a ReservationClaim? Looks like the only thing the chope node's on_request does is to assign a ticket ID and store the request, is it possible to combine ReservationRequest and ReservationClaim, i.e. assign a ticket ID upon a new claim request, and do away with the response?

In its current format, yes we can combine the messages into one. I think this is a result of older attempts where we were trying out a ticket based architecture. The idea is the reservation system is made ware early and has time to optimize. This is not part of the proto-reservation system. I can consolidate the two messages.

arjo129:
A bit of an update. I've been working on trying to diagnose the Patrol issue. In the mean time I've written some "Integration tests" here: https://github.com/arjo129/stress_test_rmf_chope. They do rely on the custom openrmf/rmf_demos#212 branch.

arjo129:
I'm still not 100% sure why the robots are getting lost but with regards to patrol tasks being problematic:

The use of `expected_next` when building Patrol tasks means that the reservation system is unessecary as the planner looks at a longer time horizon. This gives us 2 options:
1. Ignore `expected_next` when the chope node is enabled. Since the `chope_node` is enabled, the planner does not need to know what the `expected_next` destination is. This is slighly problematic as the chope node will assign parking spots if the other location is occupied resulting in oscillations in the planner.
2. If we recieve an `expected_next` block off all the locations in the `chope_node`. This will reduce oscillations, however the system will be much more conservative. New tasks will have to wait for the patrol to finish before the robot proceeds to the location. This also opens up the issue of  "what if expected_next locations are occupied?"

For next-gen rmf this should not be a problem as `rmf_reservations` already handles next durations correctly.

arjo129:
So I've had some time to debug the patrol tasks. It turns out that my previous comment was not valid. It turns out there
was an inconsitency within the chope node itself. The issue was that there were duplicates being added to the queue. This resulted in the Chope node deadlocking. I've addressed that with the latest comit.

The oscillations in the planner are out of scope of this PR. One way to address this would be to set directions on the navgraph lanes to prevent conflicts as it seems the swapping behaviour of the planner is a bit hit-or-miss.

xiyuoh:
Thanks for the fix! I've tried to run various scenarios and the chope node is behaving as expected. The planner oscillations are occurring fairly often especially for swap cases, though I agree that it is out of the scope of this PR.

luca-della-vedova (commented):
Bunch of minor things (again try to go through the code and review headers / CMakeLists).

I believe I managed to break it, the system doesn't seem to cope well with task cancellation. For example:

```
ros2 run rmf_demos_tasks dispatch_patrol -p hardware_2 --use_sim_time # tinyRobot2 will proceed
# When robot is halfway through, let's say close to the door
ros2 run rmf_demos_tasks cancel_task -id patrol.dispatch-[id]
# Now ask _the other_ robot to go there.
ros2 run rmf_demos_tasks dispatch_patrol -p hardware_2 --use_sim_time -F tinyRobot -R tinyRobot1
```

The reservation is _not_ cancelled when the task is cancelled, so the other robot will be stuck until the robot holding the reservation is asked to go anywhere else.

I'm also not too excited about the removal of the responsive wait in the demo scenario since it makes scenarios like the one below possible:

arjo129:
> The reservation is not cancelled when the task is cancelled, so the other robot will be stuck until the robot holding the reservation is asked to go anywhere else.

This is by design. If the reservation is cancelled, I need to know where to go next right? Presumably most deployments have a default task to fall back on. Again this is not enabled by default, perhaps this behaviour can be documented.

rmf_chope_node/src/main.cpp
```diff
@@ -0,0 +1,568 @@
...
+    _ticket_to_header.emplace(_last_issued_ticket_id, request_header);
+    _last_issued_ticket_id++;
+    return ticket;
+  }
+
+  rmf_chope_msgs::msg::Ticket get_existing_ticket(const std::size_t index)
+  {
+    rmf_chope_msgs::msg::Ticket ticket;
+    ticket.ticket_id = index;
+    ticket.header = _ticket_to_header[index];
```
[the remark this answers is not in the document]
mxgrey: I agree, buffer overflow is too dangerous to risk, even when it seems innocuous.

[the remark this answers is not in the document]
arjo129: Was ficed in f424094d9efc2bc7a685b9b3b434b8dec3d9617f

arjo129:
@luca-della-vedova I've updated the demo branch as we discussed offline, we now move all the chope node demo into a new launch file, making it clearer to the users this is opt-in behaviour.

mxgrey (changes requested):
I wound up with a lot of points of feedback, but the vast majority are very small items to cleanup some debugging cruft.

Thanks for this massive effort.. I know putting a new substantial piece of functionality into this codebase is a perilous and grueling task.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.hpp
```diff
@@ -0,0 +1,57 @@
...
+  void add_ticket(const rmf_reservation_msgs::msg::ReservationAllocation alloc)
+  {
+    allocations.push_front(alloc);
+  }
+
+  std::optional<rmf_reservation_msgs::msg::ReservationAllocation> release_ticket()
+  {
+    if (allocations.size() <= 1)
+    {
+      // For safety every robot must have at least one reservation at any point in time.
```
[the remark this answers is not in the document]
arjo129: So I went back on the patrol decision. For perform_action its a bit hairy I can think of use cases where we want to hold and ones where we don't want to hold. If the perform action is used to trigger cleaning the floor I think releasing would make sense. On the other hand if it were used for something else, like say wiping the cupboard in place then we should hold it. Also after a clean sequence do we come back to the same spot?

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.hpp
```diff
@@ -0,0 +1,57 @@
...
+      return std::nullopt;
+    }
+    auto temp = allocations.back();
+    allocations.pop_back();
```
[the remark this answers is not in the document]
arjo129: > Let's give some thought to whether it's possible to arrange the implementation that way.

This is going to be very hard cause we have to persist the reservation across tasks. I have no idea how we plan on dealing with which reservation to release. For now we are only dealing with parking spots so a robot should only be at one parking spot at any given time. I don't forsee us using this in the current generation in any other way.

rmf_reservation_node/src/main.cpp
```diff
@@ -0,0 +1,590 @@
...
+    const rmf_reservation_msgs::msg::ReleaseRequest::ConstSharedPtr& request)
+  {
+    RCLCPP_INFO(
+      this->get_logger(), "Releasing ticket for %s",
+      ticket_store_.debug_ticket(request->ticket.ticket_id).c_str());
+    auto ticket = request->ticket.ticket_id;
+    auto released_location = current_state_.release(ticket);
+    if (!released_location.has_value())
+    {
+      RCLCPP_ERROR(
```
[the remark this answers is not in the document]
arjo129: We probably should never get here, if we do something has gone really wrong. The fleet adapter and this node both need to be reset.

rmf_reservation_node/src/main.cpp
```diff
@@ -0,0 +1,590 @@
...
+      this->get_logger(), "Released ticket %lu, location %s is now free",
+      request->ticket.ticket_id,
+      released_location->c_str());
+
+    // Traverse waitgraph.
+    while (auto next_ticket =
+      queue_manager_.service_next_in_queue(released_location.value()))
+    {
+      // Release the ticket
+      released_location = current_state_.release(next_ticket.value());
```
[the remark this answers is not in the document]
arjo129: This triggers the next robot to "release" its waiting spot so that the next robot can come in to the location. I'm not sure `finalize` captures that relation. Maybe `make_space_for_next_robot`. `service_next_in_queue` could be renamed to `notify_next_robot`.

rmf_reservation_node/src/main.cpp
```diff
@@ -0,0 +1,590 @@
...
+
+    return ((hash<string>()(header.robot_name)
+      ^ (hash<string>()(header.fleet_name) << 1)) >> 1)
+      ^ (hash<uint64_t>()(header.request_id) << 1);
+  }
+};
+
+
+/// Ticket generation class for book keeping purposes. Will eventually overflow.
+/// Ticket id 0 does not exist and is useful for making emergency claims.
```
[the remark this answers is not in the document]
arjo129: Yeah tbf I think this comment is no longer relevant.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -281,6 +285,9 @@ class EasyFullControl::RobotConfiguration
   /// be used.
   std::optional<bool> responsive_wait() const;
 
+  /// Check if the robot should use the parking spot system.
+  bool use_parking_spot_system() const;
```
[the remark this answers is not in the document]
arjo129: I don't think a system that mixes robots with parking reservations and without them makes much sense at this current point. I may be wrong. This should be set globally.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -585,7 +595,8 @@ class EasyFullControl::FleetConfiguration
     bool default_responsive_wait = false,
     double default_max_merge_waypoint_distance = 1e-3,
     double default_max_merge_lane_distance = 0.3,
-    double min_lane_length = 1e-8
+    double min_lane_length = 1e-8,
+    bool use_parking_reservation = false
```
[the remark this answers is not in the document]
arjo129: b57cb3e

rmf_fleet_adapter/src/rmf_fleet_adapter/events/EmergencyPullover.cpp
```diff
@@ -242,66 +292,148 @@ void EmergencyPullover::Active::_find_plan()
...
+
+        self->_state->update_status(Status::Underway);
+        self->_state->update_log().info(
+          "Found a plan to move from ["
+          + start_name + "] to [" + goal_name + "]");
+
+
+        std::vector<rmf_traffic::agv::Plan::Goal> no_future_projections;
+        auto full_itinerary = project_itinerary(
+          *result, no_future_projections,
```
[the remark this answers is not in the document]
arjo129: 3e5ccc2

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
```diff
@@ -190,60 +190,69 @@ class ReservationNodeNegotiator :
...
+            {
+              negotiator->_retry_timer->cancel();
+            }
+            else
+            {
+              return;
+            }
+            for (std::size_t i = 0; i < negotiator->_goals.size(); ++i)
+            {
+              if (negotiator->_goals[i].waypoint() == current_location[0].waypoint())
```
mxgrey, line 249: The current location is not a reliable source of truth on what point is reserved for a robot.

This should be checking if the a goal waypoint is already held by the [`context->_reservation_mgr`](https://github.com/open-rmf/rmf_ros2/blob/9e16a6e7985f97432c7a01c0ffe29b305a66b368/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.hpp#L990).

arjo129: Thanks for spotting it. I've reworked this part and the reservation manager itself.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.hpp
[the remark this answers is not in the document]
arjo129: Reworked this to be much cleaner.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.hpp
```diff
@@ -24,33 +24,26 @@
...
 
 class ReservationManager
 {
 public:
-  void add_ticket(const rmf_reservation_msgs::msg::ReservationAllocation alloc)
-  {
-    allocations.push_front(alloc);
-  }
+  /// Adds a ticket and releases the previous ticket if the ticket id is different
+  void add_ticket(
```
mxgrey, line 33: Very minor nitpick: `add` gives the impression that this will be combined with the previous ticket. What if we call this `replace_ticket` instead?

rmf_reservation_node/src/main.cpp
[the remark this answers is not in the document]
mxgrey: Okay I misunderstood the purpose then. I thought `next_ticket` would represent the reservation for the next robot that is being served `released_location`, so we wouldn't be releasing that ticket now. But now I understand that `next_ticket` represents the location where the next robot had been waiting.

I wonder if renaming it to something like `previous_waiting_location` would make the logic more clear. It's okay either way, but I suspect if anyone looks over this code in the future they might experience the same confusion that I did.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
```diff
@@ -0,0 +1,382 @@
...
+            }
+            else
+            {
+              return;
+            }
+            for (std::size_t i = 0; i < negotiator->_goals.size(); ++i)
+            {
+              if (events::wp_name(*negotiator->_context.get(), negotiator->_goals[i]) == negotiator->_context->_get_reserved_location())
+              {
+                RCLCPP_ERROR(negotiator->_context->node()->get_logger(),
```
mxgrey, line 219: This looks like it should be INFO, not ERROR.

Also to be a useful log message I think we should include both the robot name and the location that was identified as already reserved.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
```diff
@@ -0,0 +1,382 @@
...
+                "Already at goal no need to engage reservation system\n");
+                negotiator->_selected_final_destination_cb(negotiator->_goals[i].waypoint());
+                return;
+              }
+            }
+            RCLCPP_INFO(negotiator->_context->node()->get_logger(),
+            "Sending reservation request");
```
mxgrey, line 226: Let's include the robot name and the location that's being reserved in order to make this log message useful.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
```diff
@@ -0,0 +1,382 @@
...
+        );
+      }
+
+      for (std::size_t i = 0; i < negotiator->_goals.size(); ++i)
+      {
+        if (events::wp_name(*context.get(), negotiator->_goals[i]) == context->_get_reserved_location())
+        {
+          RCLCPP_ERROR(context->node()->get_logger(),
+            "Already at goal no need to engage reservation system\n");
+          negotiator->_selected_final_destination_cb(negotiator->_goals[i]);
```
mxgrey, line 238: This block is redundant with the one inside the retry timer.

I actually think it would make the most sense to run this check before we check `current_location` and outside of the worker schedule callback. Then if the check passes we would do
```
context->worker()->schedule([
    cb = negotiator->_selected_final_destination_cb,
    wp = negotiator->_goals[i]
  ](const auto&)
  {
    cb(wp);
  });
return;
```
Then we don't need to reimplement this block twice.

arjo129: I relaise we don't need thew current location any more cause we uwse the reservation manmager to get the previous ticket. I'll clean up this block.

xiyuoh (commented):
Thanks for all the progress, this is looking really great!

Here are some observations from further stress testing:

**1. Office world, using the `office_reservation_node.launch.xml`**

At startup, with each robot at the respective chargers, I sent `tinyRobot2` to `tinyRobot1_charger`. This leads `tinyRobot2` to go to `tinyRobot1_charger`, even though it is already occupied. Are waypoints claimed and allocated on launch? I suspect this has something to do with the `nothing` finishing request for this fleet. This only happens when the first task submitted on launch leads a robot to a waypoint currently occupied by another robot.

**2. To test out the reservation node in multi-level scenarios, I tested it in Hotel world.**

Tweaks to building map, fleet configs and launch file:
I added a parking spot to an existing waypoint on L3, enabled the reservation node in Hotel world's launch file, and in the fleet configs set `use_reservation_node` to `True` for both `deliveryRobot` and `tinyRobot` fleets. I also set the finishing request to `nothing` for both fleets.

The reservation system works well as expected when sending robots from different fleets to the same waypoint on the same map 🎉 

There is a particular multi-level scenario I was curious about. With both `tinyBot_1` and `deliveryBot_1` at L1 at their respective chargers, I followed these steps:
- Submit patrol task for `tinyBot_1` to `L3_master_suite`. The robot begins its way up to L3.
- Submit patrol task for `deliveryBot_1` to `L3_master_suite`. The robot stays put at `deliverybot_charger`.
- After `tinyBot_1` has reached `L3_master_suite`, submit patrol task for `tinyBot_1` to `L3_room1`.
   **Observation**: `tinyBot_1` moves to `L3_room1`. At the same time at L1, `deliveryBot_1` begins moving towards the lift to travel up to L3, then goes to `L3_master_suite`.

I assume this is the intended behavior, where if the target waypoint is on another level and is occupied, the robot would go to a parking spot on its current map while waiting for the target waypoint to become vacant.
However, I was thinking it might make more sense for `deliveryBot_1` to wait at a parking spot on L3 (if one exists) instead of at L1 where it originally was. That way, if `L3_master_suite` becomes vacant while `deliveryBot_1` is making its way up to the L3 parking spot, the robot would already be halfway to the target waypoint.
With the current behavior, the travel time for each robot to travel to a different map becomes exclusive: `tinyBot_1` completes the travel from L1 to L3, waits X minutes at `L3_master_suite` and leaves, then `deliveryBot_1` starts travelling from L1 to L3. As opposed to having both robots start moving from L1 to L3 once they receive their tasks, and their travel times would overlap.
I can imagine a significant amount of waiting time saved if this was a larger building, and each robot was planned to occupy the target waypoint for few minutes to perform some action, while the travel time from charging spots to target waypoints could take anywhere from a quarter to half an hour.

This is more of a _what-should-be-the-best-behavior_ consideration, feel free to share your thoughts!

arjo129:
> Are waypoints claimed and allocated on launch? I suspect this has something to do with the nothing finishing request for this fleet.

Sadly, in the current architecture of RMF this is actually very hard to do. There must be some thing that asks the robot to go to a location to register it. Currently, if a robot starts on a location, unless there is a task assigned to it at start up, the robot will not issue a claim. If you notice in my example for the office world. I disable `finishing_request` on both. I suspect we need to be explicit about this in the documentation.

> There is a particular multi-level scenario I was curious about. With both tinyBot_1 and deliveryBot_1 at L1 at their respective chargers, I followed these steps

This is tricky, as we define `GoToPlace` as having multiple possible end points. If multiple end points are specified it would be hard to know which parking spot is nearest. Currently, the heuristic used is the nearest parking spot from the start point since there may be more than one end point. Perhaps in the event that there is only one end point I should use a different cost function.

rmf_reservation_node/src/main.cpp
```diff
@@ -0,0 +1,641 @@
...
+
+    // If we can't proceed immediately add the ticket to a queue.
+    std::stringstream ss;
+    ss << "Could not immediately service %lu, enqueing. Locations in ticket were:" <<
+      request->ticket.ticket_id << std::endl;
+    for (std::size_t i = 0; i < requests_[request->ticket.ticket_id].size();
+      i++)
+    {
+      ss << "\t- %s" << requests_[request->ticket.ticket_id][i].location.c_str() << std::endl;
+    }
```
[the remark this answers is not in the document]
arjo129: [3b69cca](https://github.com/open-rmf/rmf_ros2/pull/325/commits/3b69ccae6fe711f6acb94749623539e9f761f2b5)

arjo129:
As discussed offline, I'm implementing @xiyuoh's recommendation of waiting near the goal destination if we have only one goal destination.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
```diff
@@ -0,0 +1,367 @@
...
+      RCLCPP_ERROR(
+        _context->node()->get_logger(),
+        "reservation: Robot [%s] can't get location",
+        _context->requester_id().c_str());
+      return;
+    }
+
+    RCLCPP_INFO(
+      _context->node()->get_logger(),
+      "reservationvationvation Negotiator: Selecting a new go_to_place location from [%lu] choices for robot [%s]",
```
xiyuoh, line 263: Typo

xiyuoh (commented):
Thanks Arjo! Tested both single and multi level scenarios with a single goal waypoint, in both cases the robot waiting for the goal to be freed up was able to travel to the parking spot nearest to the goal. Other than the minor typo everything LGTM, happy to approve after that is fixed 😄

xiyuoh (dismissed):

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ChopeNegotiator.hpp
mxgrey: There's some redundancy, but the context for each use case is different enough that I don't think we stand to gain much by funneling it all into one function. Each has slight nuances that we would need to cram into the arguments of the single function. I think we should leave this as-is and be more mindful of hygiene in general as we develop the next generation.

mxgrey (approved):
I've done a final pass and I'm very happy with how this feature has turned out.

@arjo129 good job cramming this very important feature into a structure that wasn't originally made for it.

@luca-della-vedova thanks for the very thorough initial review!

@xiyuoh thanks for the rigorous testing and catching so many important edge cases!

This was an excellent team effort all around. I'll merge this within the next 24 hours as long as there are no objections.

PR #327, opened 2024-02-22
Add Backward-ROS for improved logging in event of segfaults

arjo129:
### Implemented feature

[`backward-ros`](https://github.com/pal-robotics/backward_ros/) provides a clean way to catch unhandled exceptions and segfaults and is able to print a stack trace when a program crashes. This PR adds `backward-ros` to rmf ros2 nodes so that we can trace causes for segfaults.

Note: This will only work for pure C++ nodes. If we want logging in downstream fleet adapters they will have to handle that themselves either using `backward-cpp` or `faulthandler` in python.

### DIFF
--- a/rmf_fleet_adapter/package.xml
+++ b/rmf_fleet_adapter/package.xml
@@ -36,6 +36,7 @@
   <depend>rmf_websocket</depend>
   <depend>nlohmann-json-dev</depend>
   <depend>nlohmann_json_schema_validator_vendor</depend>
+  <depend>backward_ros</depend>
 
   <build_depend>eigen</build_depend>
   <build_depend>yaml-cpp</build_depend>
--- a/rmf_fleet_adapter/package.xml
+++ b/rmf_fleet_adapter/package.xml
@@ -12,31 +12,33 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <depend>rclcpp</depend>
+
+  <depend>backward_ros</depend>
+  <depend>nlohmann_json_schema_validator_vendor</depend>
+  <depend>nlohmann-json-dev</depend>
   <depend>rclcpp_components</depend>
-  <depend>rmf_utils</depend>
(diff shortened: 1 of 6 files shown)


rmf_fleet_adapter/package.xml
```diff
@@ -36,6 +36,7 @@
   <depend>rmf_websocket</depend>
   <depend>nlohmann-json-dev</depend>
   <depend>nlohmann_json_schema_validator_vendor</depend>
+  <depend>backward_ros</depend>
```
Yadunund, line 39: Nit: Sort alphabetical for all package.xmls changed.

Yadunund (changes requested):
Thanks for adding this functionality.

I'm not very familiar with `backward_ros/backward-cpp` and have the following questions:
* Is it used in any other large projects in the ROS/Gazebo ecosystem?
* Is there any additional overhead from its usage?
* Any stats on increase in compile times if any?
slightly concerned about the stability of the pkg since it is externally maintained. One nice thing is that it is released into Rolling from `ros2-gbp` so it should be available for each new ROS distro which is nice.

arjo129:
> Is it used in any other large projects in the ROS/Gazebo ecosystem?

According to https://index.ros.org/p/backward_ros/#iron-deps seems like a lot of `ros2_controls` packages are using it.

> Is there any additional overhead from its usage?

My understanding is that no. This package simply installs a signal handler for `SIGSEGV`. The nice thing is that if we build with RelWithDebugInfo it should spit out the right stack trace thus speeding up debugging by a lot.

>Any stats on increase in compile times if any?

Nothing noticeable for me. I suspect a minor increase as we would have to link a new library in - but its not like we are linking templatized code.

Yadunund:
Thanks for the info. Sounds great to include! We can merge after addressing the sorting of dep pkgs.

Yadunund (approved):

PR #328, opened 2024-02-28
Support labels in booking information

aaronchongth:
### Implemented feature

Requires https://github.com/open-rmf/rmf_task/pull/110

Copy labels from request json into task booking information


mxgrey (approved):

mxgrey:
I think the build issue on rolling is an intermittent build farm problem, so let's go ahead and merge this.

PR #330, opened 2024-03-12
Disable automatic retreat

xiyuoh:
This PR aims to
- Make `retreat_to_charger` configurable from the fleet config, and
- Prevents `retreat_to_charger` from being triggered if the robot's finishing task is already a Charge task

We ran into scenarios where the robot is on its way back to the charger (as part of the finishing task) while its battery level is nearing the recharge threshold. This sometimes triggers a retreat to charger that overrides the finishing task. As a result, the robot is unable to perform other tasks queued later on until its current battery SOC reaches the recharge SOC.

In the `retreat_to_charger` callback, we now check the fleet's configured behavior before initiating automatic retreat.

This depends on https://github.com/open-rmf/rmf_task/pull/114


rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
```diff
@@ -706,6 +712,12 @@ class EasyFullControl::FleetConfiguration
   /// Set whether or not to account for battery drain during task planning.
   void set_account_for_battery_drain(bool value);
 
+  /// Get whether or not to trigger automatic retreat to charger.
+  bool retreat_to_charger() const;
+
+  /// Set whether or not to trigger automatic retreat to charger.
+  void set_retreat_to_charger(bool value);
+
```
Yadunund, line 720: I don't think these APIs should be added here. Instead they should be present in `FleetUpdateHandle` instead. Users who rely on `EasyFullControl` API can get a `FleetUpdateHandler` handler using the [EasyFullControl::more() API](https://github.com/open-rmf/rmf_ros2/blob/a2269abafa827195690e6e83d9b2a1b9d552d3f8/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp#L134-L135) and set this config.

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -1628,6 +1628,10 @@ void TaskManager::retreat_to_charger()
 
   if (!task_planner->configuration().constraints().drain_battery())
     return;
+  if (!task_planner->configuration().constraints().retreat_to_charger())
+    return;
+  if (_idle_task && _idle_task->request_type() == "Charge")
```
Yadunund, line 1633: We shouldn't have comparison to magic stings like this. The `rmf_task::Request` class should have a virtual static member function that should return the request type. Then you can call `_idle_task && _idle_task->request_type() == ParkRobotIndefinitely::request_type()`

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp
```diff
@@ -308,7 +313,8 @@ class FleetUpdateHandle : public std::enable_shared_from_this<FleetUpdateHandle>
     double recharge_threshold,
     double recharge_soc,
     bool account_for_battery_drain,
-    rmf_task::ConstRequestFactoryPtr finishing_request = nullptr);
+    rmf_task::ConstRequestFactoryPtr finishing_request = nullptr,
+    bool retreat_to_charger = true);
```
Yadunund, line 317: I don't think this should be a bool. Instead it should be the timer period. If this is set to `std::nullopt`, we reset the timer thereby disabling the automatic retreat to charger behavior. This way we not only expose the ability to enable/disable this feature but also allow users to configure the rate at which the check happens. Else we will need to add another parameter for that.

The docstring above should also be updated and it's worth mentioning the timer there so users know how this check is performed.

```suggestion
    std::optional<rmf_traffic::Duration> check_retreat_to_charger_interval = rmf_traffic::time::from_seconds(10));
```

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
Yadunund: See comment below on setting/getting this value as an `std::optional<rmf_traffic::Duration>`.

Yadunund (changes requested):
Thanks for working on the fix but I feel the implementation should be as follows:

1. `FleetUpdateHandle::set_task_planner_params()` should accept an `std::optional<rmf_traffic::Duration>` param.
2. Add a `TaskManager::retreat_to_charger(std::optional<rmf_traffic::Duration> duration)` method. If a value is provided this method will execute https://github.com/open-rmf/rmf_ros2/blob/a2269abafa827195690e6e83d9b2a1b9d552d3f8/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp#L184-L192. If `std::nullopt`, it will call `mgr->_retreat_timer.reset()` to disable the timer.
3. Setting the task planner params or using the new setter in `FleetUpdateHandle` will invoke the method in 2. This would allow for more efficient processing as if the check is not needed, the timer callback will not execute thereby saving some cycles for the ROS 2 executor.

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
Yadunund: With the proposed refactor, we no longer rely on checking request types.

xiyuoh:
Updated according to the changes requested, without the check for the fleet's configured idle task before deciding whether to proceed with the automatic retreat to charger. I'll work on that in a separate PR.

mxgrey (approved):
I've made some changes to prevent the API from breaking and for a few other nitpicky reasons.

I've left in-line comments as part of this review to explain the changes that are being made, but no further action is needed.

mxgrey (approved):

PR #333, opened 2024-03-14
Fix interaction between emergency pullover and finishing task

mxgrey:
When an emergency pullover interrupts a finishing task, the robots that were doing the finishing task would not resume the finishing task correctly.

With this PR, finishing tasks will resume correctly after an emergency pullover is released.


xiyuoh (dismissed):
Tested this with a `charge` finishing task and it works great! The finishing task is triggered as expected.

xiyuoh (approved):

PR #334, opened 2024-03-15
Release other mutexes if robot started charging

xiyuoh:
This PR aims to release any mutex groups a robot may be holding once it starts charging at its dedicated charging waypoint. This is to ensure that mutex groups won't be stuck with a stationary robot for an indefinite period of time, blocking other robots from moving into these mutexes.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/WaitForCharge.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/WaitForCharge.cpp
@@ -94,6 +94,19 @@ WaitForCharge::Active::Active(
     .get_observable()
     .start_with(initial_msg);
 
+  // If the charging waypoint has a mutex group, release all other mutexes
+  std::unordered_set<std::string> retain_mutexes;
+  const auto charging_waypoint = _context->dedicated_charging_wp();
+  const auto& graph = _context->navigation_graph();
+  retain_mutexes.insert(
+    graph.get_waypoint(charging_waypoint).in_mutex_group());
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/phases/WaitForCharge.cpp
```diff
@@ -94,6 +94,19 @@ WaitForCharge::Active::Active(
...
     .start_with(initial_msg);
 
+  // If the charging waypoint has a mutex group, release all other mutexes
+  std::unordered_set<std::string> retain_mutexes;
+  const auto charging_waypoint = _context->dedicated_charging_wp();
+  const auto& graph = _context->navigation_graph();
+  retain_mutexes.insert(
+    graph.get_waypoint(charging_waypoint).in_mutex_group());
+  _context->retain_mutex_groups(retain_mutexes);
+  RCLCPP_INFO(
```
mxgrey, line 104: Minor nitpick to make this message more useful. Let's change it to something like this:
```
if (retain_mutexes.front().empty())
{
  RCLCPP_INFO(
    _context->node()->get_logger(),
    "Robot [%s] is waiting to charge. All its mutex groups will be released.",
    _context->name().c_str());
}
else
{
  RCLCPP_INFO(
    _context->node()->get_logger(),
    "Robot [%s] is waiting to charge. It will retain only the mutex group [%s].",
    _context->name().c_str(),
    retain_mutexes.front().c_str());
}
```
Giving more specific information can make confusing situations easier to debug.

mxgrey (approved):

PR #338, opened 2024-03-28
Stabilize commissioning feature

mxgrey:
We have had a decommission feature in the unstable API for a while now, but it left some loose ends, e.g. what should be done with pending tasks that were already assigned to the decommissioned robot?

This PR adds a stable API for changing the commissioning of a robot. It provides a number of settings that allow the commission to be customized based on the needs of the deployment. Acceptance of dispatched tasks, direct tasks, and performing idle behavior can all be toggled separately. Any of these three can be toggled on or off at any time.

When requesting a change in commission it is also possible to specify what should happen with pending tasks for the robot. For dispatched tasks, there are three options, with `reassign` being the default:
* `reassign`: The robot's tasks will be reassigned to other robots within the fleet. If the robot is being decommissioned for dispatched tasks then its pending tasks will be distributed across all the remaining robots in the fleet that are currently commissioned for dispatch requests. If there are no more robots in the fleet that are commissioned for dispatch requests, then the pending tasks will be canceled. If `reassign` is used while **re**commissioning a robot (toggling its `dispatch_tasks` commission to `true`) then the recommissioned robot may immediately draw in pending tasks from other robots in the fleet.
* `complete`: There will be no change to the pending dispatched tasks of the robot. Even if the `dispatch_tasks` commission is being toggled off, the robot will continue to complete its pending task queue but will not accept any new dispatched tasks.
* `cancel`: All pending dispatched tasks for the robot will be canceled. If set, this will take effect whether the robot's `dispatch_tasks` commission is being toggled on or off or remaining unchanged, so avoid using this setting recklessly. 

The only options provided for handling the pending direct tasks are `complete` and `cancel` with `complete` being the default. We currently do not provide a way to reassign pending direct tasks from one robot to another.

Any changes to commissioning do not impact any tasks that are currently running. RMF does not currently provide a reliable way to hand off an ongoing task from one robot to another. If a robot that is currently running a task has been decommissioned, then the operator or system integrator must decide whether to allow the ongoing task to finish or to additionally issue a request to cancel the ongoing task. This choice will likely depend on operational and situational details, which is why this API does not take any automatic action on it.

This PR also fixes some previously existing issues:
* If a robot in the fleet happens to begin one of its pending tasks in the very brief window between when a new task bid began and when that bid gets awarded, then that new task would never actually be run. This was a bug that has been overlooked because this race condition has such an extremely rare window that we have never noticed it happen before.
* In some rare cases a newly awarded task bid would require the fleet adapter to replan all the tasks for its fleet, and that replanning was taking place on the main event loop worker, which would potentially delay all other main event loop callbacks that need to be processed. This has been fixed so that all task planning happens on a separate worker.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -2009,6 +2009,8 @@ rmf_task::State TaskManager::_publish_pending_task(
 {
   const auto info = pending.request()->description()->generate_info(
     std::move(expected_state), parameters);
+  PendingInfo cache;
+  cache.info = info;
 
   nlohmann::json pending_json;
   const auto& booking = *pending.request()->booking();
@@ -2022,13 +2024,18 @@ rmf_task::State TaskManager::_publish_pending_task(
 
   if (pending.finish_state().time())
   {
-    pending_json["unix_millis_finish_time"] =
+    PendingTimeInfo t;
+    t.unix_millis_finish_time =
       to_millis(pending.finish_state().time()->time_since_epoch()).count();
 
     const auto estimate =
       pending.finish_state().time().value() - pending.deployment_time();
-    pending_json["original_estimate_millis"] =
-      std::max(0l, to_millis(estimate).count());
+    t.original_estimate_millis = std::max(0l, to_millis(estimate).count());
+
+    pending_json["unix_millis_finish_time"] = t.unix_millis_finish_time;
+    pending_json["original_estimate_millis"] = t.original_estimate_millis;
+
+    cache.time = t;
   }
   copy_assignment(pending_json["assigned_to"], *_context);
   pending_json["status"] = "queued";
@@ -2041,6 +2048,7 @@ rmf_task::State TaskManager::_publish_pending_task(
 
   _validate_and_publish_websocket(task_state_update, validator);
 
+  _pending_task_info[pending.request()] = cache;
   return pending.finish_state();
 }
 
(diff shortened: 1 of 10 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -1043,7 +1147,11 @@ nlohmann::json TaskManager::submit_direct_request(
       }
       catch (const std::exception&)
       {
-        json_errors.push_back(e);
+        nlohmann::json error;
+        error["code"] = 42;
```
aaronchongth, line 1151: any chance this is supposed to be 422?

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -829,52 +853,39 @@ void FleetUpdateHandle::Implementation::dispatch_command_cb(
...
+              "TaskAssignments updated for robots in fleet [%s].",
+              task_id.c_str(), name.c_str());
+          },
+          // On failure
+          [task_id, name = name, node = node](std::vector<std::string> errors)
           {
-            ss << "The following planner errors occurred:";
-            for (const auto& e : errors)
+            std::stringstream ss;
+            ss << "Unabled to replan assignments when cancelling task ["
```
aaronchongth, line 870:
```suggestion
            ss << "Unable to replan assignments when cancelling task ["
```

aaronchongth (approved):
Very small nitpicks/questions, but everything else looks good on the first pass, thanks!

PR #341, opened 2024-04-16
Emergency trigger to use transient local qos

aaronchongth:
This makes the emergency topic subscriber transient local.


arjo129 (commented):
A quick grep shows that the[ LiftSupervisor node](https://github.com/open-rmf/rmf_ros2/blob/b3113fd2468991a91501ae9710cd9cfe77caaef5/rmf_fleet_adapter/src/lift_supervisor/Node.cpp#L52) also publishes to this topic with default QoS. I'm not sure if we're using that but might be worth updating.

mxgrey:
We [don't actually publish](https://github.com/open-rmf/rmf_ros2/blob/b3113fd2468991a91501ae9710cd9cfe77caaef5/rmf_fleet_adapter/src/lift_supervisor/Node.cpp#L139) on the emergency topic from the lift supervisor. A long time ago the intention was to do so, but it was later decided that the lift states are not necessarily a good source of truth for this, so now it's up to system integrators to decide when to publish emergency notices. 

If anything we should delete that publisher from the lift supervisor.

arjo129 (approved):

PR #342, opened 2024-04-22
Fix deadlock in websocket server

arjo129:
## Bug fix

OK So I'm not too happy with this "Fix" but it probably will resolve our deadlock issue.

This fixes a deadlock we were facing in production where in the event the server fails to connect. It is rather strange that the underlying library does not trigger an `Failed` or `Closed` event. Perhaps I should set a [Pong Timeout handler](https://github.com/zaphoyd/websocketpp/blob/1b11fd301531e6df35a6107c1e8665b1e77a2d8e/websocketpp/endpoint.hpp#L302)


aaronchongth (approved):
LGTM

PR #346, opened 2024-04-24
Automatically begin or cancel idle behavior when commission changes

mxgrey:
It was recently noticed that idle behavior will not be automatically cancelled when it gets toggled off during a commissioning change. It seems extremely unlikely that there will ever be a situation where someone wants the idle behavior to be decommissioned but does not want to cancel it. Also idle behavior is usually "hidden" so operators often don't have a way to explicitly ask for it to be cancelled.

This PR fixes the situation by automatically cancelling idle behavior if there is any idle behavior happening when idle behavior gets toggled off in a robot's commissioning. This PR also ensures that the idle behavior will begin automatically if the idle behavior gets toggled back on in a robot's commissioning.


aaronchongth (approved):
Thanks for handling this behavior! LGTM

PR #348, opened 2024-05-06
Add reassign task allocation API

xiyuoh:
This PR exposes the `reassign_dispatched_tasks` API to be usable from fleet adapter implementations. This allows users to request for a replan of task allocation at runtime in scenarios, especially in scenarios where a robot may be stuck on a task longer than expected with a queued task.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
@@ -2539,6 +2539,13 @@ FleetUpdateHandle& FleetUpdateHandle::set_retreat_to_charger_interval(
 }
 
 //==============================================================================
+FleetUpdateHandle& FleetUpdateHandle::reassign_dispatched_tasks()
+{
+  _pimpl->reassign_dispatched_tasks([]() {}, [](auto) {});
+  return *this;
+}
+
+//==============================================================================
 std::shared_ptr<rclcpp::Node> FleetUpdateHandle::node()
 {
   return _pimpl->node;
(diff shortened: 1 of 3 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -2538,6 +2538,13 @@ FleetUpdateHandle& FleetUpdateHandle::set_retreat_to_charger_interval(
   return *this;
 }
 
+//==============================================================================
+FleetUpdateHandle& FleetUpdateHandle::reassign_dispatched_tasks()
+{
+  _pimpl->reassign_dispatched_tasks([]() {}, [](auto) {});
```
mxgrey, line 2544: This needs to be scheduled inside of the fleet handle's worker, similar to what's done for the [RobotUpdateHandle version](https://github.com/open-rmf/rmf_ros2/blob/132e516d5c7e55486cd30ba422ee09a1e648fc1e/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp#L807-L813) of this function. Neglecting to do that can lead to segmentation faults and undefined behavior.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -2538,6 +2538,13 @@ FleetUpdateHandle& FleetUpdateHandle::set_retreat_to_charger_interval(
   return *this;
 }
 
+//==============================================================================
+FleetUpdateHandle& FleetUpdateHandle::reassign_dispatched_tasks()
```
mxgrey, line 2542: Returning `FleetUpdateHandle&` is generally done in situations where users may want to chain a series of function calls. Usually that's when you're changing some settings for a class.

This function is triggering a distinct action, so I don't think it makes sense stylistically to chain any more function calls after it. Let's just return `void` from this function.

mxgrey (changes requested):

xiyuoh:
Thanks for the review @mxgrey , comments are addressed in [7a1b24c](https://github.com/open-rmf/rmf_ros2/pull/348/commits/7a1b24c65a4603444ea96ad181d5ed1ac09d3c14)

PR #356, opened 2024-05-24
Cherry pick linking time fix to Humble

luca-della-vedova:
Cherry pick of #297 to `humble` branch.
This came up as part of https://github.com/open-rmf/rmf_ci_templates/pull/10, moving from `lld` to `ld` in our action caused the `humble` branches that don't have the fix to timeout on linking.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,6 +13,9 @@ jobs:
     name: rmf_ros2
     uses: open-rmf/rmf_ci_templates/.github/workflows/reusable_build.yaml@main
     with:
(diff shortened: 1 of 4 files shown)


luca-della-vedova:
While CI is running, note that I fully expect `rolling` / `jammy` to fail going forward, I'm happy to remove it if we don't want the noise

mxgrey (approved):

mxgrey:
> I'm happy to remove it if we don't want the noise

I think it makes perfect sense to remove rolling for this branch. We don't care how the humble branch does against rolling anyway.

PR #357, opened 2024-05-27
Quiet cancel API

xiyuoh:
This PR adds a `quiet_cancel` feature to the TaskManager. This feature allows the fleet adapter to end tasks early from the RobotUpdateHandle. The underlying logic is equivalent to a task being cancelled, but the task phases and status are overwritten to be displayed as `Completed`. The TaskManager now has a `_quiet_cancel` flag to indicate whether we are handling a normal or quiet cancel request.

This API is added to the RobotUpdateHandle Unstable API.


mxgrey (approved):
Looks good, thanks for putting this together!

PR #358, opened 2024-05-31
Add icecream as a rosdep dependency

arjo129:
Turns out this was missing from the `package.xml`. Since we use it we should include it as a `rosdep` and add a rosdep key. 

Depends on https://github.com/ros/rosdistro/pull/41432


Yadunund (approved):
LGTM. Will merge after `main` is open again.

Issue #252, opened 2022-11-24
Cannot use timestamp to get task_id 


allstarry:
### Required information:

- Operating system and version:
  -  ubuntu 22.04
- OpenRMF installation type:
  -  build from source 
- OpenRMF version or commit hash
  -  latest on main
- ROS distribution and version:
  -  humble
- ROS installation type:
  -  binary 
- Package or library, if applicable:
  - 

### Description of the bug

Set `use_timestamp_for_task_id` to true and `use_sim_time` to false, then the task_id will always be `dispatch--2147483648`

### Steps to reproduce the bug

1. Set `use_timestamp_for_task_id` to true and `use_sim_time` to false, then build 
1. Launch RMF
1. Send a task request
1. See the error displayed

### Expected behavior

Get an correct and unique task_id from response

### Actual behavior

Get wrong task_id from response and it doesn't change when a new task is requested

### Additional information

Launch file is shown below

```
<?xml version='1.0' ?>

<launch>
  <arg name="use_sim_time" default="false"/>
  <arg name="failover_mode" default="false"/>
  <arg name="initial_map" default="default" description="Initial map name for the visualizer"/>
          <!--  <arg name="viz_config_file" default="$(find-pkg-share rmf_visualization_schedule)/config/rmf.rviz"/> -->
  <arg name="viz_config_file" default ="$(find-pkg-share rmf_launch)/config/office.rviz"/>
  <arg name="headless" default="true" description="do not launch rviz; launch gazebo in headless mode"/>
  <arg name="bidding_time_window" description="Time window in seconds for task bidding process" default="2.0"/>
  <arg name="use_rmf_panel" default="true" description="launch web and api server for rmf_demos_panel"/>
  <arg name="dashboard_config_file" default="$(find-pkg-share rmf_demos_dashboard_resources)/office/dashboard_config.json"/>
  <arg name="use_timestamp_for_task_id" description="" default="true"/>


  <!-- TinyRobot fleet adapter and robot state aggregator needed for the TinyRobot slotcar_plugin -->
  <group>
    <let name="fleet_name" value="AMR"/>
    <include file="$(find-pkg-share rmf_launch)/launch/agv.launch.xml">
      <arg name="fleet_name" value="$(var fleet_name)"/>
      <arg name="use_sim_time" value="$(var use_sim_time)"/>
      <arg name="nav_graph_file" value="/home/bayes/amr_ws/routemap/nav_graphs/0.yaml" />
    </include>
[...]

PR #366, opened 2024-06-11
Fix bug when using timestamps for dispatcher task IDs

aaronchongth:
### Fixed bug

Fixes https://github.com/open-rmf/rmf_ros2/issues/252. This issue only shows up when the dispatcher node is not running with sim time due to casting issues when the double gets larger.

To recreate the issue, set `use_sim_time` to `false` when launching the dispatcher node in 
https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos/launch/common.launch.xml#L60

### Fix applied

* using `std::lround` instead of `std::round)`


luca-della-vedova (approved):

mxgrey:
We should probably wait for approval from @Yadunund before merging so we don't disturb the release process that's going on.

Yadunund:
Yep kindly hold for a day or two more. Shall update here.

luca-della-vedova:
> Yep kindly hold for a day or two more. Shall update here.

I think we should be OK right?

PR #373, opened 2024-06-25
Populate LiftProperties for Destination in localize callbacks

xiyuoh:
This PR populates the LiftProperties stored in a `Destination` object when `localize()` is being called. This ensures that the lift data can be retrieved within the callback.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/RequestLift.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/RequestLift.cpp
@@ -235,18 +235,9 @@ void RequestLift::ActivePhase::_init_obs()
             agv::Destination::Implementation::get(*me->_data.localize_after)
             .position = me->_context->position();
 
-            rmf_traffic::agv::Graph::LiftPropertiesPtr in_lift;
             const auto graph = me->_context->navigation_graph();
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/phases/RequestLift.cpp
```diff
@@ -235,6 +235,19 @@ void RequestLift::ActivePhase::_init_obs()
             agv::Destination::Implementation::get(*me->_data.localize_after)
             .position = me->_context->position();
 
+            rmf_traffic::agv::Graph::LiftPropertiesPtr in_lift;
+            const auto graph = me->_context->navigation_graph();
+            for (const auto& lift : graph.all_known_lifts())
```
mxgrey, line 240: You should be able to use `graph.find_known_lift(lift->name())` instead of using this for-loop. Let me know if you run into any issue with that.

mxgrey (changes requested):

rmf_fleet_adapter/src/rmf_fleet_adapter/phases/RequestLift.cpp
xiyuoh: Ah thanks for pointing me to this API! It works great

mxgrey (approved):

PR #374, opened 2024-06-28
Use main branch for repos file in CI

luca-della-vedova:
Some deployments might need newer features but still rely on older ROS distros (i.e. Humble), this PR changes the CI to always use the latest `main` branches for CI and test it against a few different underlying ROS distros (i.e. Humble, Iron, Jazzy and Rolling).

Note to reviewer, the order of operations should be:

* Merge the upstream PR https://github.com/open-rmf/rmf_ci_templates/pull/15
* Revert the change on this PR that uses a custom branch on `rmf_ci_templates`.
* Do a CI run
* Merge this PR

### DIFF
--- a/.github/workflows/asan.yaml
+++ b/.github/workflows/asan.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   asan_test:
     name: rmf_ros2 asan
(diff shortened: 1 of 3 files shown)


mxgrey (approved):
Thanks for getting this to work!

PR #376, opened 2024-07-15
Provide an API that says the robot's lift destination

mxgrey:
This PR addresses https://github.com/open-rmf/rmf/discussions/496 by providing an API in `RobotUpdateHandle` that will contain the final destination of a lift that the robot is about to use.

By the time a robot receives a command to enter the lift, the `RobotUpdateHandle::lift_destination()` function will contain information about where the lift will be going. If the robot is not interacting with a lift, it will return a `std::nullopt`.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
@@ -855,6 +855,8 @@ RobotUpdateHandle::lift_destination() const
   {
     return context->final_lift_destination();
   }
+
+  return std::nullopt;
 }
 
 //==============================================================================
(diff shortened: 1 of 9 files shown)


luca-della-vedova (approved):
Tested with hotel world by adding the new API call and checking that it returns the right values when robots take lifts, looks good!

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotUpdateHandle.cpp
```diff
@@ -828,6 +828,35 @@ void RobotUpdateHandle::reassign_dispatched_tasks()
...
+}
+
+//==============================================================================
+std::optional<RobotUpdateHandle::LiftDestination>
+RobotUpdateHandle::lift_destination() const
+{
+  if (const auto context = _pimpl->get_context())
+  {
+    return context->final_lift_destination();
+  }
```
cwrx777, line 857: shouldn't there be a return when there is no context?

mxgrey: Great catch, fixed here: [ca7d865](https://github.com/open-rmf/rmf_ros2/pull/376/commits/ca7d86593b3ae9c5f899fab4dc2103861a080a38)

I'm very annoyed that the C++ compiler doesn't prevent this by default 😮‍💨

luca-della-vedova (approved):

PR #377, opened 2024-07-15
Fixes warning given by `colcon_ros` on Jazzy

arjo129:
This is a small fix for when we build systems on jazzy we get the warning:

```
[5.452s] WARNING:colcon.colcon_ros.task.ament_python.build:Package 'rmf_charging_schedule' doesn't explicitly install the 'package.xml' file (colcon-ros currently does it implicitly but that fallback will be removed in the future)
```

This pr fixes this warning by installing the `package.xml`


luca-della-vedova (approved):
Noticed the warning as well, thanks for addressing it

PR #383, opened 2024-08-22
Publish fleet and task updates over ROS 2 if websocket is not provided

aaronchongth:
### Implemented feature

Allows `fleet_state_update`, `fleet_log_update`, `task_state_update`, and `task_log_update` to be published over ROS 2 when websocket server URL is not provided.

Works with https://github.com/open-rmf/rmf-web/pull/1003

The motivation is related to https://github.com/open-rmf/rmf-web/pull/899, where non-reproducible issues with websockets occur unpredictably in production environments, causing various systems to fail.

This allows the option to mitigate issues with the upstream `websocketpp` library, and just rely on ROS 2 to publish updates instead.

This will also allow the use of ROS 2 introspection tools in production if network instability occurs, which hopefully is a much better experience than debugging websocket connections.

## Testing

Instead of running demos with the `server_uri`, just run

```bash
ros2 run rmf_demos_gz office.launch.xml #headless:=1
```

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -193,7 +193,7 @@ TaskManagerPtr TaskManager::make(
     });
 
   auto reliable_transient_qos =
-    rclcpp::ServicesQoS().keep_last(20).transient_local();
+    rclcpp::ServicesQoS().keep_last(10).reliable().transient_local();
   mgr->_task_state_update_pub =
     mgr->_context->node()->create_publisher<TaskStateUpdateMsg>(
       TaskStateUpdateTopicName, reliable_transient_qos);
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -199,7 +199,7 @@ TaskManagerPtr TaskManager::make(
       TaskStateUpdateTopicName, reliable_transient_qos);
   mgr->_task_log_update_pub =
     mgr->_context->node()->create_publisher<TaskLogUpdateMsg>(
-      TaskLogUpdateTopicName, reliable_transient_qos);
+      TaskLogUpdateTopicName, reliable_transient_qos.keep_last(100));
 
   const std::vector<nlohmann::json> schemas = {
     rmf_api_msgs::schemas::task_state,
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -192,6 +192,18 @@ TaskManagerPtr TaskManager::make(
         self->_handle_request(request->json_msg, request->request_id);
     });
 
+  if (!mgr->_broadcast_client.has_value())
+  {
+    auto reliable_transient_qos =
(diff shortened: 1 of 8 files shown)


koonpeng:
Can it publish regardless if websocket is enable?

Connection issues aside, I feel that the split of multiple transport is causing unnecessary complexities as well. websockets is the "public" api but it doesn't expose many important information (like door, lifts, maps etc), the ros2 side is "internal", does expose them but it doesn't expose task states etc. So in practice, it is very likely for downstream to have to listen to both transports and deal with 2 different message schema format.

rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
```diff
@@ -294,9 +300,15 @@ class Dispatcher::Implementation
         this->handle_dispatch_ack(*msg);
       });
 
+    task_state_update_pub = node->create_publisher<TaskStateUpdateMsg>(
+      rmf_task_ros2::TaskStateUpdateTopicName,
+      rclcpp::ServicesQoS().keep_last(10).reliable().transient_local());
```
koonpeng, line 305: Is there a reason why this is using a different QoS from https://github.com/open-rmf/rmf_ros2/pull/383/files#diff-979ff9700f77b8e1260de3301125242adb32556ba5092fcddcc39e91baa09a2dR195-R196?

koonpeng (approved):

rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp
aaronchongth: thanks for spotting that, I've corrected them to be the same https://github.com/open-rmf/rmf_ros2/pull/383/commits/07b9c6abc23200c30123bf681c2ceedc62458dba

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -192,6 +192,15 @@ TaskManagerPtr TaskManager::make(
         self->_handle_request(request->json_msg, request->request_id);
     });
 
+  auto reliable_transient_qos =
+    rclcpp::ServicesQoS().keep_last(10).reliable().transient_local();
```
mxgrey, line 196: Since this is going to be used for logs which could experience sudden bursts that we don't want to miss, I would suggest having at least `.keep_last(100)` for logs. But since task states always send out the latest state, replacing whatever was sent before, 10 would be sufficient.

Maybe we should have a different history value for each.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -531,6 +542,13 @@ class FleetUpdateHandle::Implementation
         });
     }
 
+    handle->_pimpl->fleet_state_update_pub =
+      handle->_pimpl->node->create_publisher<FleetStateUpdateMsg>(
+        FleetStateUpdateTopicName, reliable_transient_qos);
+    handle->_pimpl->fleet_log_update_pub =
+      handle->_pimpl->node->create_publisher<FleetLogUpdateMsg>(
+        FleetLogUpdateTopicName, reliable_transient_qos);
```
mxgrey, line 550: I recommend using 100 history for this log topic as well.

mxgrey (changes requested):
Just one very minor recommendation on QoS history length for log topics, and then I'll be happy to approve.

mxgrey (approved):
Looks good!

PR #384, opened 2024-08-23
Update _last_active_task_time if robot is executing task

cwrx777:
### Fixed bug
Reference: open-rmf/rmf#516

### Fix applied
Set `_last_active_task_time` to `now` if the robot is currently doing task.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
@@ -1496,15 +1496,17 @@ void RobotContext::_check_door_supervisor(
 {
   const auto now = std::chrono::steady_clock::now();
   const auto dt = std::chrono::seconds(10);
-
   if (_current_task_id.has_value())
   {
     _last_active_task_time = now;
   }
-  else if (_last_active_task_time + dt < now)
+  else
   {
-    // Do not hold a door if a robot is idle for more than 10 seconds
-    _holding_door = std::nullopt;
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
```diff
@@ -1496,7 +1496,12 @@ void RobotContext::_check_door_supervisor(
 {
   const auto now = std::chrono::steady_clock::now();
   const auto dt = std::chrono::seconds(10);
-  if (_last_active_task_time + dt < now)
+
```
Yadunund, line 1499: Nit: Unnecessary newline.
```suggestion
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.cpp
```diff
@@ -1496,7 +1496,12 @@ void RobotContext::_check_door_supervisor(
...
+  if (_current_task_id.has_value())
+  {
+    _last_active_task_time = now;
+  }
+  else if (_last_active_task_time + dt < now)
```
Yadunund, line 1504: Some style guides may complain if we have an `else if` block without an `else` so better if we can include something like.
```
else
{
  continue;
}

```

Yadunund (changes requested):

Yadunund (approved):
LGTM from me. Will wait for an approval from @mxgrey before merging.

mxgrey (approved):
Makes sense, thanks for the fix!

PR #385, opened 2024-09-11
set request_time in lift end session request.

cwrx777:
This PR is just to set `request_time` for `REQUEST_END_SESSION` lift request.


luca-della-vedova (approved):
Thanks!

PR #392, opened 2024-10-30
Allow automatic action cancellation to be toggled

mxgrey:
Historically the way task cancellation worked for "perform action" events was to immediately declare that the action is cancelled and trust the action implementer to somehow gracefully wind down the action even though we might immediately start issuing new commands.

With this PR, the action implementer can use `execution.set_automatic_cancel(false)` to prevent the task system from immediately declaring that the cancellation is finished. Instead the task system will wait until the implementer calls `execution.finished()`, at which point the cancellation will be registered.

If the action implementer does not call `execution.set_automatic_cancel(false)` then the old behavior will remain as it has always been.


xiyuoh (approved):
Thanks for the fix! This API addition is great and can be easily applied on any execution object, tried it out in demos with Office world's teleop task too and it works perfectly.

PR #393, opened 2024-11-11
Fix lift disruption issue

mxgrey:
This PR fixes https://github.com/open-rmf/rmf_ros2/issues/389

The cause of the problem was a subtle error in some conditional logic which only gets triggered when a task gets cancelled or a replan occurs while a robot is waiting for a lift but the lift has not yet arrived. Beginning in https://github.com/open-rmf/rmf_ros2/pull/310 the horizon of a robot is only reported up to the point where it reaches a lift and will need to wait for the lift to arrive. Once the lift arrives, the robot will replan to produce the next plan horizon that it will report to the traffic schedule. There was a flawed conditional statement inside of RequestLift to watch for when this occurs and then short-circuit the RequestLift logic.

That conditional statement is fixed by having RequestLift preserve the arrived / not arrived state inside the RobotContext and then checking that before it tries to short-circuit.

### DIFF
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -38,6 +38,7 @@ set(dep_pkgs
   rmf_battery
   rmf_task
   rmf_task_sequence
+  rmf_reservation_msgs
   std_msgs
   rmf_api_msgs
   rmf_websocket
@@ -88,6 +89,7 @@ ament_target_dependencies(rmf_fleet_adapter
     rmf_dispenser_msgs
     rmf_ingestor_msgs
     rmf_building_map_msgs
+    rmf_reservation_msgs
     rclcpp
 )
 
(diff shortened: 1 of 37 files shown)


mxgrey:
> _lift_arrived will only be set to true when RequestLift::ActivePhase::_finish() is called, which will only happen after the has_lift_arrived() check at the start of _init_obs().

Unfortunately it's a very confusing and circular workflow, but notice that [`_context->request_replan()`](https://github.com/open-rmf/rmf_ros2/blob/f028518b56a308de7d91ebd052a506d3b410335d/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/RequestLift.cpp#L472) gets triggered after `_set_lift_arrived` is used. This will cause the plan to cycle back and redo the `RequestLift`, at which point we want to short-circuit the logic so we don't get thrown back into that cycle.

This strange cycle is done because waiting for the lift can take so long that the remainder of the plan might need to change by the time the lift arrives. Once we know that the lift has arrived and ready for the robot to board, we do this replan based on the latest information about what's going on in the environment before sending the robot into the lift.

For the next generation we should do something much cleaner with workflows, but we're stuck with this weird cyclical logic for now.

xiyuoh (approved):
I see, thanks for explaining! The fix LGTM, tested it a few times in a modified hotel world as well and the bug seems to be gone 👍

PR #398, opened 2024-12-09
Do not skip waypoint if orientation is not aligned

xiyuoh:
This PR is a follow-up to https://github.com/open-rmf/rmf_traffic/pull/111 targeting issue https://github.com/open-rmf/rmf/issues/556.

When a new path is provided to EasyFullControl from the traffic planner, `follow_new_path` would look for a connection between the path waypoints and the robot's current location in order to determine an overlapping starting point. EFC would then skip any unnecessary waypoints before the starting point to prevent the robot from backtracking. However, if the fleet config's `skip_rotation_commands` is set to False, these waypoints should not be skipped if they are responsible for in-place rotation.

In the proposed fix, if `skip_rotation_commands` flag is set to False, it would check for the orientation difference between the robot and potential starting point. It ensures in-place rotation waypoints are not skipped if the difference is too large (threshold is measured at 0.01, taken from another portion of the same code [here](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp#L1255)).

For example, if the following path is provided, with the robot's current pose `[18.8078, -10.3719, 0.0306]` and `skip_rotation_commands` set to False:
```
[fleet_adapter-16] -- [wp_0] 18.7944, -10.3724, 0.0304936
[fleet_adapter-16] -- [wp_1] 18.7944, -10.3724, 1.58649
[fleet_adapter-16] -- [wp_2] 18.7395, -6.87398, 1.58649
[fleet_adapter-16] -- [wp_3] 18.7395, -6.87398, 1.49677
[fleet_adapter-16] -- [wp_4] 18.9498, -4.03908, 1.49677
[fleet_adapter-16] -- [wp_5] 18.9498, -4.03908, -0.0127633
[fleet_adapter-16] -- [wp_6] 13.1306, -3.96481, -0.0127633
[fleet_adapter-16] -- [wp_7] 13.1306, -3.96481, -0.296291
[fleet_adapter-16] -- [wp_8] 7.03417, -2.11099, -0.296291
[fleet_adapter-16] -- [wp_9] 7.03417, -2.11099, 1.14971
[fleet_adapter-16] -- [wp_10] 6.52584, -3.24596, 1.14971
```
Without the fix, the fleet adapter would skip waypoints `wp_0` and `wp_1`, and request the robot to move directly from its current pose to `wp_2`.
With the fix applied, the new path would only skip waypoint `wp_0` and commands the robot to move to `wp_1` from its current pose, which is an in-place rotation, before moving on to `wp_2`.

Another minor change includes changing `max_merge_lane_distance` to `max_merge_waypoint_distance` [here](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp#L1076), since this condition block deals with waypoints in the same stack rather than the robot being on a lane.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
@@ -1055,9 +1055,21 @@ void EasyCommandHandle::follow_new_path(
   for (std::size_t i = 0; i < waypoints.size(); ++i)
   {
     const auto& wp = waypoints[i];
-    if (wp.graph_index().has_value())
+    for (const auto& l : current_location)
     {
-      for (const auto& l : current_location)
+      if (!nav_params->skip_rotation_commands)
+      {
+        // If the waypoint and current orientations are not aligned with
+        // skip rotation command set to False, we will not consider this as
+        // a connection
+        if (std::abs(wp.position()[2] -
+          l.orientation())*180.0 / M_PI > 1e-2)
+        {
+          continue;
+        }
+      }
+
+      if (wp.graph_index().has_value())
       {
         if (nav_params->in_same_stack(*wp.graph_index(),
           l.waypoint()) && !l.lane().has_value())
@@ -1073,7 +1085,7 @@ void EasyCommandHandle::follow_new_path(
             {
               const double dist =
                 (*l.location() - wp.position().block<2, 1>(0, 0)).norm();
-              if (dist <= nav_params->max_merge_lane_distance)
+              if (dist <= nav_params->max_merge_waypoint_distance)
               {
                 found_connection = true;
                 i0 = 0;
@@ -1108,10 +1120,7 @@ void EasyCommandHandle::follow_new_path(
           }
         }
       }
-    }
-    else
-    {
-      for (const auto& l : current_location)
+      else
       {
         Eigen::Vector2d p_l;
         if (l.location().has_value())
@@ -1137,6 +1146,17 @@ void EasyCommandHandle::follow_new_path(
       {
         for (const auto& l : current_location)
         {
+          if (!nav_params->skip_rotation_commands)
+          {
+            // If the waypoint and current orientations are not aligned with
+            // skip rotation command set to False, we will not consider this as
+            // a connection
+            if (std::abs(wp.position()[2] -
+              l.orientation())*180.0 / M_PI > 1e-2)
+            {
+              continue;
+            }
+          }
           if (l.lane().has_value())
           {
             if (lane == *l.lane())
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
```diff
@@ -1055,9 +1055,21 @@ void EasyCommandHandle::follow_new_path(
...
-      for (const auto& l : current_location)
+      if (!nav_params->skip_rotation_commands)
+      {
+        // If the waypoint and robot orientations are not aligned with
+        // skip rotation command set to False, we will not consider this as
+        // a connection
+        if (std::abs(wp.position()[2] -
+          l.orientation())*180.0 / M_PI > 1e-2)
+        {
+          continue;
```
luca-della-vedova, line 1068: Note that because there are two nested for loops this `continue` will affect the inner one, which is the for loop for `current_location` and not the one that is iterating `waypoints`, is that what you meant to do? The same applies to the other snippet

xiyuoh: Yep! We'd still want to check this waypoint against other possible current locations (e.g. on lane A, or lane B, or another RMF waypoint) of the robot. While it is likely consistent, I'm not sure the locations in `current_location` would _always_ have the same orientation, so iterating through all combis to be safe.

luca-della-vedova (approved):

luca-della-vedova:
Thanks for the explanation! Does this depend on https://github.com/open-rmf/rmf_traffic/pull/111 or can it be merged independently?

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp
```diff
@@ -1055,9 +1055,21 @@ void EasyCommandHandle::follow_new_path(
...
+        }
+      }
+
+      if (wp.graph_index().has_value())
```
luca-della-vedova, line 1072: I just realized you switched the position of:

```
if (wp.graph_index().has_value())
```

and

```
for (const auto& l : current_location)
```

Since `wp` is constant throughout the loop iterations it makes sense to keep it outside of the loop and skip the whole loop if the condition is false, otherwise in cases where `wp = std::nullopt` we would be iterating and doing the `skip_rotation_commands` check just to do nothing in the loop afterwards

luca-della-vedova: Ah nevermind just realized the branching is more complex than I thought

xiyuoh:
This PR is independent, can be merged first. Thanks!

PR #400, opened 2025-01-20
Improve arrival time predictions for lanes with speed limits

mxgrey:
It was brought up in https://github.com/open-rmf/rmf/discussions/580 that traffic schedule visuals are not turning out correctly for lanes that have speed limits. I think we might not be able to make the visual work perfectly, but this PR should at least help somewhat.

### DIFF
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -517,12 +517,22 @@ rmf_api_generate_schema_headers(
 
 # -----------------------------------------------------------------------------
 
-ament_export_targets(rmf_fleet_adapter HAS_LIBRARY_TARGET)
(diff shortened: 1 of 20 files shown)


xiyuoh:
Thanks for working on a fix! Unfortunately I'm still seeing the jump in the yellow schedule marker when passing over a speed limited lane. 
speed_limit

mxgrey:
I found what I had overlooked during my first attempt. This should now be fixed by [17955e9](https://github.com/open-rmf/rmf_ros2/pull/400/commits/17955e990c7d184306137d23d483a68da7e0d6b5)

I did a test run and found that it tracks the slowed down speed very nicely.

xiyuoh (approved):
The fix worked great, LGTM!

PR #403, opened 2025-02-20
Fix potential issues around tasks starting

mxgrey:
This PR fixes two potential issues that can come up around tasks starting:
* Make the default threshold for charging the battery more reasonable. An exact 100.0% charge might never get reported by some batteries.
* Direct tasks weren't triggering the `_begin_next_task` function of the `TaskManager`; instead they relied on the ROS timer to trigger it. Sometimes the ROS timer that periodically triggers `_begin_next_task` doesn't fire off for long periods of time, causing direct tasks to take inordinately long before they begin. With this PR, we always trigger `_begin_next_task` after a direct task request has been received.


xiyuoh (approved):
LGTM!

Issue #404, opened 2025-02-21
[Bug]: deliveryBot waits unnecessarily long before passing through doors in Hotel world


xiyuoh:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 24.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

e4d1d87

### ROS distribution

Rolling

### ROS installation type

Source build

### Description of the bug

When sending `deliveryBot_1` to travel through doors in Hotel world, it would be waiting unreasonably long before passing through doors, even though the doors have been opened for some time.

When waiting, replanning/opening door logs are being printed constantly. Here's a small snippet:
```
[fleet_adapter-19] [INFO] [1740113118.970417798] [deliveryRobot_fleet_adapter]: Executing final go_to_place [restaurant] for robot [deliveryRobot/deliveryBot_1]
[fleet_adapter-19] [INFO] [1740113118.972393618] [deliveryRobot_command_handle]: Commanding [deliveryBot_1] to navigate to [ 11.46772539 -21.78732521   1.58231167] on map [L1]: cmd_id 840
[fleet_adapter-19] [INFO] [1740113119.174047294] [deliveryRobot_fleet_adapter]: Opening door [L1_door2] for [deliveryRobot/deliveryBot_1]
[fleet_adapter-19] [INFO] [1740113119.597983514] [deliveryRobot_fleet_adapter]: Requesting replan for [deliveryRobot/deliveryBot_1] because it is too far from its commanded path. Details:
[fleet_adapter-19] Current location
[fleet_adapter-19]  -- L1 < 11.4677 -21.7873> [#9] | location < 11.4578 -21.7905> | orientation 90.6598
[fleet_adapter-19] Commanded path:
[fleet_adapter-19]  -- t=0 #9 < 11.4677 -21.7873  1.58231> yaw=90.6598 event [DoorOpen L1_door2] approach lanes:
[fleet_adapter-19]  -- t=10.5217 #10 < 11.4227 -17.8757  1.58231> yaw=90.6598 event [DoorClose L1_door2] approach lanes: 18
[fleet_adapter-19] [INFO] [1740113119.598401834] [deliveryRobot_fleet_adapter]: Replanning requested for [deliveryRobot/deliveryBot_1]
[...]

PR #405, opened 2025-02-21
Fix the criteria for replanning

mxgrey:
Issue #404 reported an unnecessarily long wait when passing through a door. This PR fixes that problem.

Essentially we were using `max_merge_waypoint_distance` to judge whether a robot is close enough to where it was supposed to be when we should've been using `max_merge_lane_distance`. The "merge waypoint" distance is supposed to be much stricter than the "merge lane" distance because merging waypoints implies that the waypoints are essentially the same whereas merging lanes means the robot can move along the path as though it's somewhere on the lane.

Using the less strict parameter is now allowing the scenario to work as expected.


xiyuoh (approved):
Tested and works great, thanks for the quick fix!

PR #406, opened 2025-02-25
check if _chosen_goal has value before accessing

piliwilliam0306:
### Fixed bug

if `_chosen_goal` doesn't have value, it may raise:
```
terminate called after throwing an instance of 'std::bad_optional_access'
```
This PR checks if `_chosen_goal` has value before accessing

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
@@ -200,14 +200,6 @@ auto GoToPlace::Active::make(
         if (const auto c = self->_context->command())
           c->stop();
 
-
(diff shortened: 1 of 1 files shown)


mxgrey (commented):
Very good catch, you're right that this should not be here. I think it was a debug printout that should've been removed before it was merged into `main`.

I think we should actually remove the log entirely since `_find_plan` is already logging all relevant updates about the chosen goal.

piliwilliam0306:
> Very good catch, you're right that this should not be here. I think it was a debug printout that should've been removed before it was merged into `main`.
> 
> I think we should actually remove the log entirely instead of protecting it with an if-statement since `_find_plan` is already logging all relevant updates about the chosen goal.

@mxgrey 
Thanks for reviewing, removed!

mxgrey (approved):

PR #411, opened 2025-04-01
return if no ticket when cancel reservation

piliwilliam0306:
### Fixed bug

return if `!has_ticket()` when cancel reservation instead of `has_ticket()`

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.cpp
@@ -53,7 +53,7 @@ void ReservationManager::cancel()
   {
     return;
   }
-  if (has_ticket())
+  if (!has_ticket())
     return;
 
   RCLCPP_INFO(


arjo129 (approved):
Thank you for spotting this. Yes! This is definitely a bug which could lead to undefined behavior. I think the main reason it is not being triggered is that we don't have a test for this path. Our integration tests for the reservation system are kind of clunky and live out here: https://github.com/arjo129/stress_test_rmf_chope

Before I merge this I'd like to write a test for it to verify the behavior is correct.

piliwilliam0306:
@arjo129 
I've also noticed that https://github.com/open-rmf/rmf_ros2/blob/9f1ec091711ae476335996301f1caaefac3a02fd/rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp#L46 is not called when calling `cancel()`.
It seems this might be because the object was already destroyed, causing an early return on https://github.com/open-rmf/rmf_ros2/blob/9f1ec091711ae476335996301f1caaefac3a02fd/rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp#L43

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
```diff
@@ -33,18 +33,7 @@ class ReservationNodeNegotiator :
...
-    {
-      auto self = ptr.lock();
-      if(!self)
-      {
-        return;
-      }
-
-      self->_context->_cancel_allocated_destination();
-    });
+    _context->_cancel_allocated_destination();
```
arjo129, line 36: There is a good reason for us to check if the Negotiator exists and only run it on a worker thread. Please revert this change.

arjo129 (changes requested):

rmf_fleet_adapter/src/rmf_fleet_adapter/events/internal_ReservationNodeNegotiator.hpp
piliwilliam0306: OK, thx, reverted.

arjo129:
Thanks for this contribution @piliwilliam0306! I've merged in your commits to #416, however in hind sight since the reservation system never used the cancel (and there is no need to use cancel with the finishing task), it is better to completely remove this dead code path.

Issue #82, opened 2021-07-15
Investigate data races reported by Thread Sanitizer #46


aaronchongth:
To recreate this build, clone everything on `main` in a new workspace

```bash
export CC=clang
export CXX=clang++

cd ws
source /opt/ros/foxy/setup.bash
colcon build --mixin tsan --cmake-args -DCMAKE_BUILD_TYPE=Debug --packages-up-to rmf_fleet_adapter
```

Package tests can be run individually, for example,

```bash
source install/setup.bash
export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp

# This one seems the most prominent.
./build/rmf_fleet_adapter/test_rmf_fleet_adapter
```

Quickly skimming through it, I can see that the first topic of interest would be the constructor of `rmf_traffic_ros2::blockade::BlockadeNode`, which similarly uses the node context. The race conditions happen on both `rmw_fastrtps_cpp` and `rmw_cyclonedds_cpp`.

The log can still be downloaded from the `tsan` action in #46 
[4_tsan_build_test.txt](https://github.com/open-rmf/rmf_ros2/files/6822191/4_tsan_build_test.txt)

Some `gdb` commands that might be helpful, thanks to https://stackoverflow.com/questions/63964739/how-to-add-breakpoint-when-thread-sanitizer-reports-data-race

```bash
gdb ./build/rmf_fleet_adapter/test_rmf_fleet_adapter

(gdb) break exit
(gdb) break _exit
(gdb) catch syscall exit_group 
(gdb) set env TSAN_OPTIONS=halt_on_error=1
(gdb) run
```

aaronchongth:
Per VC, we should attempt this build on Galactic to see if any of errors are fixed.

aaronchongth:
Running the `test_rmf_fleet_adapter` in Galactic, we are getting less data races, but some are still present, https://gist.github.com/aaronchongth/7481bac63c018de0c9bcf71d720b633f

aaronchongth:
First data race in `test_TrafficLight`,
* Creation of blockade node, https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/test/adapters/test_TrafficLight.cpp#L170-L172
* Creation of mock adapter, https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/test/adapters/test_TrafficLight.cpp#L200-L202

aaronchongth:
[...]

PR #413, opened 2025-04-04
Giving options to separate fire alarm by fleet names

TanJunKiat:
Our current deployment uses one RMF instance for two buildings and when one fire alarm is triggered, the robots from another building will also go into emergency mode which is not ideal.

This PR, together with the PR [#82](https://github.com/open-rmf/rmf_internal_msgs/pull/82) in `rmf_internal_msgs`, aims to resolve this.

The fire alarm message will contain the fleet names which it wants to alert and the fleet adapter will filter to see if the signal is for them.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
@@ -42,6 +42,7 @@
 #include <rmf_fleet_msgs/msg/dock_summary.hpp>
 #include <rmf_fleet_msgs/msg/lane_states.hpp>
 #include <rmf_fleet_msgs/msg/charging_assignments.hpp>
+#include <rmf_fleet_msgs/msg/emergency_signal.hpp>
 
 #include <rmf_fleet_adapter/agv/FleetUpdateHandle.hpp>
 #include <rmf_fleet_adapter/StandardNames.hpp>
@@ -410,7 +411,7 @@ public:
       {
         if (const auto self = w.lock())
         {
-          self->_pimpl->handle_emergency(msg->data);
+          self->_pimpl->handle_emergency(msg);
         }
       });
     handle->_pimpl->emergency_planner =
@@ -686,7 +687,7 @@ public:
 
   void update_fleet_state() const;
   void update_fleet_logs() const;
-  void handle_emergency(bool is_emergency);
+  void handle_emergency(std::shared_ptr<rmf_fleet_msgs::msg::EmergencySignal> is_emergency);
   void update_emergency_planner();
 
   void update_charging_assignments(const ChargingAssignments& assignments);
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
@@ -43,6 +43,7 @@
 #include <rmf_fleet_msgs/msg/lane_states.hpp>
 #include <rmf_fleet_msgs/msg/charging_assignments.hpp>
 #include <rmf_fleet_msgs/msg/emergency_signal.hpp>
+#include <std_msgs/msg/bool.hpp>
 
 #include <rmf_fleet_adapter/agv/FleetUpdateHandle.hpp>
 #include <rmf_fleet_adapter/StandardNames.hpp>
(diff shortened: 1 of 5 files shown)


mxgrey:
One concern that I have about this PR is it quietly breaks backwards compatibility in a way that will likely confuse existing users.

The emergency topic is named [`fire_alarm_trigger`](https://github.com/open-rmf/rmf_ros2/blob/9f1ec091711ae476335996301f1caaefac3a02fd/rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp#L79). Current users will publish a `std_msgs::msg::bool` to that topic, and I believe they would just experience a quiet discovery failure if they use this updated branch. The robots simply won't respond to the message and the user won't have any clear indication of why.

I would recommend we introduce a new topic name for this new message type. I don't have a strong opinion on what the name should be, but maybe something like

```cpp
const std::string PartialEmergencyTopicName = "partial_emergency";
```
or
```cpp
const std::string TargetedEmergencyTopicName = "targeted_emergency";
```

We should keep a subscriber for the `fire_alarm_trigger` topic and just convert the bool into an `EmergencySignal` message before passing it along to the new implementation.

TanJunKiat:
@mxgrey 

I have done as suggested and tested with the hotel simulation.

TanJunKiat:
Regarding the legacy fire alarm signal topic, should we add a warning printout to start re-directing people to use the new topic?

```
[warn] This topic, `fire_alarm_trigger` pipeline will be deprecated soon. Please use the new topic `emergency_signal` with the new topic message format.
```

While i understand your point about quiet discovery failure, it will be unwise to have 2 entry point for emergency signals. It will be harder to manage in both code and functionality.

What do you think?

TanJunKiat:
> I think we're very close on this, but I left a few inline comments on things I'd suggest tweaking.
> 
> Additionally it would be good to open a PR in the `rmf_demos` repo with a script that makes it easy to publish the emergency signal. `ros topic pub` could be used, but I don't like formatting a ROS message from the command line.
> 
> I'd suggest the script look something like
> 
> ```
> ros2 run rmf_demos_tasks emergency_signal true --fleets FleetA FleetB
> ```
> 
> so the general pattern for the arguments would be
> 
> ```
> [true|false] [--fleets A B ...]
> ```
> 
> where the `[true|false]` is a mandatory positional argument that can be interpreted as a bool and then `--fleets` is an optional argument that's followed by a list of fleet names.

Please refer to https://github.com/open-rmf/rmf_demos/pull/288 and see if it works as you suggested.

mxgrey:
> Regarding the legacy fire alarm signal topic, should we add a warning printout to start re-directing people to use the new topic?

If we use `RCLCPP_WARN_ONCE` then this seems fine to me.

If we want to move towards `emergency_signal` being **the** entry point for emergency signals then maybe we should rename

```cpp
const std::string TargetEmergencyTopicName = "emergency_signal";
```

to

```cpp
const std::string EmergencySignalTopicName = "emergency_signal";
```

It's a bit harder to distinguish it from the existing `EmergencyTopicName` global variable name, but we could mark that one as [deprecated](https://oopscenities.net/2022/01/20/cpp14-deprecated/).

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp
```diff
@@ -410,9 +413,24 @@ class FleetUpdateHandle::Implementation
       {
         if (const auto self = w.lock())
         {
+          RCLCPP_WARN_ONCE(
+            self->_pimpl->node->get_logger(),
+            "This topic is getting deprecated soon. Please use the topic 'emergency_signal instead.");
+          [[deprecated("Use target_emergency_sub instead")]]
```
mxgrey, line 419: I'm not sure what putting the deprecated tag here does. Usually it goes on a function or variable declaration, and then anyone who uses the function or variable will get a compiler warning that they are using something deprecated.

TanJunKiat: How about [here](https://github.com/open-rmf/rmf_ros2/blob/9f1ec091711ae476335996301f1caaefac3a02fd/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.hpp#L109)?

At the definition of `emergency_notice()`.

mxgrey: When I suggested using the `[[deprecated]]` it was to put it on the [`EmergencyTopicName`](https://github.com/open-rmf/rmf_ros2/pull/413/files#diff-3539bd42d2f6815e811b52bf6957ae43156d6f813669de7c37c3958f8cc69c9eR79) variable, like this:

```cpp
[[deprecated("Use EmergencySignalTopicName with the rmf_fleet_msgs/EmergencySignal message instead")]]
const std::string EmergencyTopicName = "fire_alarm_trigger";
```

That way anyone who is using `EmergencyTopicName` downstream will see a compilation warning about using it.

mxgrey (approved):
Everything is green! Thanks for the contribution @TanJunKiat !

TanJunKiat:
> Everything is green! Thanks for the contribution @TanJunKiat !

Thanks for the quick review and approval. This will help our deployment greatly.

PR #414, opened 2025-04-11
Ensure correct docking orientation

xiyuoh:
This PR targets https://github.com/open-rmf/rmf/discussions/661#discussioncomment-12772383 and adds a rotation command before the docking command in the event the robot is not facing the target docking station and `skip_rotation_commands` is set to False.


mxgrey:
While you're touching on this part of the code, I wonder if we can also address this issue: https://github.com/open-rmf/rmf/discussions/663#discussioncomment-12775066

Issue #415, opened 2025-04-21
[Bug]: Lift supervisor publishing request even when lift is at destination floor


TanJunKiat:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 24.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

main

### ROS distribution

Jazzy

### ROS installation type

Source build

### Description of the bug

The current logic of the lift supervisor is that it will keep publishing even when the lift states (destination level and/or door state) arethe desired states.

https://github.com/open-rmf/rmf_ros2/blob/41e9f914342153b7010badd896d0a96579e92f67/rmf_fleet_adapter/src/lift_supervisor/Node.cpp#L102-L110

The behavior seems off as the request should only publish when there is a need to.

Is there a pair of missing brackets to encapsulate the `if` condition?

### Steps to reproduce the bug

nil

### Expected behavior

Lift request to only publish when lift states does not match desired states.

### Actual behavior

Lift request publishing even when lift states matches desired states.

mxgrey:
I think this is _somewhat_ related to this discussion around excessive lift supervisor activity: https://github.com/open-rmf/rmf/discussions/638 so I think it's worth linking the two together.

For the specific case that you're bringing up here, in the early days of RMF we published messages pretty aggressively, even if we had good reason to believe that the message had already been published, because we regularly ran into issues where DDS fails to deliver messages that we are depending on, even if we used reliable QoS.

If users are seeing that there's too much network traffic from this aggressive publishing, then we can consider an alternative strategy, but it may come at the cost of the current guarantee of [eventual consistency](https://en.wikipedia.org/wiki/Eventual_consistency), or it may require a more complex data flow that includes an ACK protocol.

TanJunKiat:
> For the specific case that you're bringing up here, in the early days of RMF we published messages pretty aggressively, even if we had good reason to believe that the message had already been published, because we regularly ran into issues where DDS fails to deliver messages that we are depending on, even if we used reliable QoS.

Thinking out loud, does changing the `lift_requests` to a ROS2 service help with the DDS issue, since services requires a hand-shake between client and server.

Each lift adapter will be a server with a unique namepsace and the lift supervisor will request accordingly.

mxgrey:
[...]

PR #418, opened 2025-04-22
Reduce churn in lift supervisor

mxgrey:
This PR is meant to address https://github.com/open-rmf/rmf_ros2/issues/415 and https://github.com/open-rmf/rmf/discussions/638

Mostly this should reduce excessive logging by only logging activity where we expect something to change.

This might also help reduce unnecessary publishing somewhat, but probably not very much. We're filtering out lift commands using

```
    if ((lift_request->destination_floor != state->destination_floor) ||
      (lift_request->door_state != state->door_state))
```

which means if the door isn't currently in the requested state then we publish the command again. A lift door will be closed far more often than it's open, and it a lift request will generally ask for the door to be open, so I expect this won't filter very much of the lift request publishing. We could improve this if the lift state included a field for the current **_target_** door state, but that would be introducing a new field for lift adapters to account for, and I don't want to do that without significant community engagement.

I've run some tests in the hotel world and the lifts appear to still be working correctly, but those tests aren't very comprehensive. Given how crucial it is to keep lift behavior correct, I would appreciate help testing this PR on heavy traffic scenarios @TanJunKiat @cwrx777 . A test on hardware would be especially good since it would indicate that the behavior of a real hardware lift adapter has not been negatively impacted.

### DIFF
--- a/rmf_fleet_adapter/src/lift_supervisor/Node.cpp
+++ b/rmf_fleet_adapter/src/lift_supervisor/Node.cpp
@@ -61,6 +61,8 @@ void Node::_adapter_lift_request_update(LiftRequest::UniquePtr msg)
     {
       if (msg->request_type != LiftRequest::REQUEST_END_SESSION)
       {
+        // Set the timestamp to be the same so it doesn't confuse the comparison.
+        msg->request_time = curr_request->request_time;
         if (*curr_request != *msg)
         {
           RCLCPP_INFO(
--- a/rmf_fleet_adapter/src/lift_supervisor/Node.cpp
+++ b/rmf_fleet_adapter/src/lift_supervisor/Node.cpp
@@ -70,6 +70,8 @@ void Node::_adapter_lift_request_update(LiftRequest::UniquePtr msg)
           );
         }
         curr_request = std::move(msg);
+        curr_request->request_time = this->now();
+        _lift_request_pub->publish(*curr_request);
       }
       else
       {
@@ -94,6 +96,8 @@ void Node::_adapter_lift_request_update(LiftRequest::UniquePtr msg)
         msg->session_id.c_str(), msg->destination_floor.c_str(), msg->request_type
       );
       curr_request = std::move(msg);
+      curr_request->request_time = this->now();
+      _lift_request_pub->publish(*curr_request);
     }
   }
 
@@ -111,13 +115,15 @@ void Node::_lift_state_update(LiftState::UniquePtr state)
     if ((lift_request->destination_floor != state->destination_floor) ||
       (lift_request->door_state != state->door_state))
     {
+      lift_request->request_time = this->now();
       _lift_request_pub->publish(*lift_request);
     }
   }
-  else
+  else if (!state->session_id.empty())
   {
-    // If there are no active sessions going on, we keep publishing session
-    // end requests to ensure that the lift is released
+    // If the lift state has an active session but there are not supposed to be
+    // any active sessions going on, we keep publishing session end requests to
+    // ensure that the lift gets released
     LiftRequest request;
     request.lift_name = state->lift_name;
     request.destination_floor = state->current_floor;
--- a/rmf_fleet_adapter/src/lift_supervisor/Node.cpp
+++ b/rmf_fleet_adapter/src/lift_supervisor/Node.cpp
@@ -113,7 +113,8 @@ void Node::_lift_state_update(LiftState::UniquePtr state)
   if (lift_request)
   {
     if ((lift_request->destination_floor != state->destination_floor) ||
-      (lift_request->door_state != state->door_state))
+      (lift_request->door_state != state->door_state) ||
+      (lift_request->session_id != state->session_id))
     {
       lift_request->request_time = this->now();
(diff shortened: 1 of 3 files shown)


mxgrey:
@TanJunKiat @cwrx777 I'd like to get these changes into the kilted release. For that, it would be best to have it merged by this week.

If I don't get any feedback by Wednesday then I'll move forward with merging this to get it released. Let me know if either of you would like time to review it and/or test it out, and I can delay the merge a bit.

rmf_fleet_adapter/src/lift_supervisor/Node.cpp
```diff
@@ -92,31 +101,27 @@ void Node::_adapter_lift_request_update(LiftRequest::UniquePtr msg)
...
-    if ((lift_request->destination_floor != msg->current_floor) ||
-      (lift_request->door_state != msg->door_state))
-      lift_request->request_time = this->now();
-    _lift_request_pub->publish(*lift_request);
-    RCLCPP_INFO(
-      this->get_logger(),
-      "[%s] Published lift request to [%s] from lift supervisor",
-      msg->session_id.c_str(), lift_request->destination_floor.c_str()
-    );
+    if ((lift_request->destination_floor != state->destination_floor) ||
```
cwrx777, line 111: do we also need to check if the mode in the lift request is the same as in the state?

mxgrey: Good point, although I think the mode has a risk of being an unreliable indicator across different lifts. We anticipate that some lifts might not have different discrete modes, so their lift adapters might not broadcast the mode in a way that's consistent with our expectations.

Instead [70eddc8](https://github.com/open-rmf/rmf_ros2/pull/418/commits/70eddc8903a560b6cde2c86d358ad2a768c18557) checks to make sure that the `session_id` matches between that state and the current request. We do expect the lift adapters to broadcast the correct session ID once a session has been locked in for a request.

TanJunKiat:
@mxgrey 

I don't understand the intention for the publishing at these two lines, L74, and L79
https://github.com/open-rmf/rmf_ros2/blob/91609f63b9aff050adb1c5a61abfccc8f87f84f4/rmf_fleet_adapter/src/lift_supervisor/Node.cpp#L72-L85


The `LiftRequest` comes in and if valid, refresh the unodered map, and publishes to the lift adapter.

At the same time, another subscriber node checks if the level, door, and session ID is valid before publishing at every instance of `LiftState` to the lift adapter.

The first node seems redundant to me and might be the reason for the churning.

If publshing at the first node is neccessary, then can we implement the same conditional checks to reduce the publishing?

mxgrey:
> The first node seems redundant to me and might be the reason for the churning.

The first subscriber publishes as soon as a new request lands so we're not as sensitive to the lift adapter's state publishing behavior before publishing the new request. The second subscriber checks if the lift state is consistent with what the supervisor wants it to be and then sends a corrective request if it's not.

Indeed we could check the incoming request against the currently active request to filter out redundant request publishing.

TanJunKiat:
@mxgrey 

https://github.com/open-rmf/rmf_ros2/blob/91609f63b9aff050adb1c5a61abfccc8f87f84f4/rmf_fleet_adapter/src/lift_supervisor/Node.cpp#L115

Does it make more sense to check if the`lift_request ->destination_level` matches the `state->current_level`?

For our lift and lift adapter, once the lift reaches the destination, the lift server will not have a `destination` variable because the task is considered completed.

TanJunKiat:
> > The first node seems redundant to me and might be the reason for the churning.
> 
> The first subscriber publishes as soon as a new request lands so we're not as sensitive to the lift adapter's state publishing behavior before publishing the new request. The second subscriber checks if the lift state is consistent with what the supervisor wants it to be and then sends a corrective request if it's not.
> 
> Indeed we could check the incoming request against the currently active request to filter out redundant request publishing.

Hmmm understand where you are coming from. Then let's just leave the publisher in and add the conditional checks for the publisher?

mxgrey:
> For our lift and lift adapter, once the lift reaches the destination, the lift server will not have a destination variable because the task is considered completed.

I see, thanks for sharing this. This kind of consideration is one of the things that makes it tricky to implement these protocols correctly: Both sides must have the same idea about what these fields mean and when to change them.

In [4f0ee98](https://github.com/open-rmf/rmf_ros2/pull/418/commits/4f0ee98b44890b8b10d3889739bd63ddde314e02) I've made it account for the behavior that you've described.

> Then let's just leave the publisher in and add the conditional checks for the publisher?

I realized that we're already checking `if (*curr_request != *msg)` before publishing incoming requests, but that is likely to fail because of differences in the time stamp of each.

In [db3602f](https://github.com/open-rmf/rmf_ros2/pull/418/commits/db3602f380fce3436132e7fb4100562e288788c1) I'm modifying the incoming timestamp of the request before comparing the message to the currently active request, which should allow the equality check to filter the way we want.

TanJunKiat:
@mxgrey 

I am unable to test the new commits today, so if you need to close the PR today for Kilted, you can go ahead.

I will test the commits on Monday and open an issue if there is any.

But it looks good to me.

PR #420, opened 2025-05-06
Allow creating Auctioneer with node interfaces to support lifecycle nodes

aaronchongth:
### Implemented feature

* added a new builder function that takes in node interfaces
* added missing exports for `rmf_task_ros2`

So far I've verified that existing demos work as expected. Will change this to ready-for-review once I've tested downstream packages (namely `nexus_transporter::rmf_transporter`)


rmf_task_ros2/include/rmf_task_ros2/bidding/Auctioneer.hpp
```diff
@@ -80,6 +89,47 @@ class Auctioneer : public std::enable_shared_from_this<Auctioneer>
...
+    const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
+    node_logging_interface,
+    const rclcpp::node_interfaces::NodeTimersInterface::SharedPtr
+    node_timers_interface,
+    const rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr
+    node_topics_interface,
+    const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr
+    node_parameters_interface,
+    BiddingResultCallback result_callback,
+    ConstEvaluatorPtr evaluator);
```
Yadunund, line 131: Suggestion for a cleaner API. If we don't need all the node interfaces we can replace `ALL_RCLCPP_NODE_INTERFACES` with a list of interfaces we need.

```suggestion
  using ActioneerInterfaces = rclcpp::node_interfaces::NodeInterfaces<ALL_RCLCPP_NODE_INTERFACES>
  
  static std::shared_ptr<Auctioneer> make(
     ActioneerInterfaces node_interfaces,
    BiddingResultCallback result_callback,
    ConstEvaluatorPtr evaluator);
```

aaronchongth: Sweet! Yeah we don't need all of them, I've already listed out the ones that we need

rmf_task_ros2/src/rmf_task_ros2/bidding/internal_Auctioneer.hpp
```diff
@@ -34,7 +34,15 @@ class Auctioneer::Implementation
...
 
-  std::shared_ptr<rclcpp::Node> node;
+  rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_interface;
+  rclcpp::node_interfaces::NodeClockInterface::SharedPtr node_clock_interface;
+  rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
+    node_logging_interface;
+  rclcpp::node_interfaces::NodeTimersInterface::SharedPtr node_timers_interface;
+  rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr node_topics_interface;
+  rclcpp::node_interfaces::NodeParametersInterface::SharedPtr
+  node_parameters_interface;
```
Yadunund, line 44: We could just move the node_interfaces here and call the getters whenever we need access to them? (save memory footprint)

```suggestion
  ActioneerInterfaces node_interfaces;
```

rmf_task_ros2/src/rmf_task_ros2/bidding/Auctioneer.cpp
```diff
@@ -15,17 +15,21 @@
  *
 */
 
+#include <rclcpp/create_publisher.hpp>
+#include <rclcpp/create_subscription.hpp>
+#include <rclcpp/create_timer.hpp>
```
Yadunund, line 20: Can probably replace these and the `#include <rclcpp/node.hpp>` in `Auctioneer.hpp` with a single `#include <rclcpp/rclcpp.hpp>` in `Auctioneer.hpp` but keeping a minimal set of imports is fine too 👍🏼

Yadunund (approved):
LGTM

luca-della-vedova (changes requested):
This breaks compilation on `humble`, we should be absolutely sure there is no way around it and be very careful about it

mxgrey:
> I strongly believe users who want to work with humble should build off the humble branch.

The bulk of our user base is still running deployments off of humble. Due to security policies that are in their deployment contracts, they generally cannot choose to move to a newer distro. At the same time, their deployments need some of the new features that we are producing on main. Those new features often introduce new APIs or introduce significant changes to behavior, so it would violate buildfarm policy for us to backport them to an already released distro.

The combination of those factors means it's very important for the bulk of our active user base that the core Open-RMF libraries remain compatible with humble for the foreseeable future.

luca-della-vedova (commented):
The fact that all the core RMF still build with humble is helpful for users that are still on humble itself but can afford to build RMF from source. I feel it can be convenient especially when you have deployments where updating your whole OS can be a much bigger effort than updating a commit hash to try a new feature.
I'm not 100% opposed to breaking humble support but as I mentioned I feel it should be a very conscious decision, if there are new features on a new ROS distro that would significantly improve the user experience then I'm open to breaking it, but in this case it seems a fairly niche new API that, in my opinion, doesn't make it very worth it.

Yadunund:
I see. Thanks for updating my memory on the deployment user base still using Humble. 

We could backport new features to the `humble` branch especially if we're not breaking any API. In fact, even if the changes in `main` were to break API, we can backport the capabilities to older distros but by updating the implementation to ensure no API/ABI breaks. In this case new API that uses `NodeInterfaces` can be backported to `kilted` & `jazzy` while the `humble` backport would be manually modified to rely on `NodeBaseInterfaces`. 

But if there are new messages, and other API breaking changes on `main` that the deployment user base on `humble` is relying on which cannot be backported to the `humble` branch, then I understand the rationale.

Anyways, I don't mean to hold this PR back.

mxgrey:
> We could backport new features to the humble branch especially if we're not breaking any API.

My single biggest concern is behavioral changes. I'm afraid of doing a rug pull on anyone who didn't consciously choose to migrate to a newer version of the libraries. While I generally believe that all behavioral changes have been universally positive improvements, there's always a risk that someone [depends](https://xkcd.com/1172/) on the existing behavior.

I'm personally open to reconsidering this approach. We do get requests from users for the new features to be released on earlier distros. But my long-standing impression of the build farm policy has always been that new features, new APIs, and changes to behavior are all strongly discouraged on released distros, even if they're not strictly forbidden.

mxgrey:
I have a very silly idea for how we can get this PR through without needing to make a very thorny decision about release policy.

We can define a pure abstract class that wraps this set of APIs. We can define a child class that stores a `std::shared_ptr<Node>` and calls the `Node` class APIs. Then we can use an `#ifdef` to identify if we're compiling against rclcpp from jazzy or later. Inside that >= jazzy block, we can define a child class that stores a `std::shared_ptr<LifecycleNode>` and calls the `LifecycleNode` class APIs.

It's a bit tedious, but it would at least unblock the ability to use lifecycle nodes in recent distros, while still supporting humble.

aaronchongth:
thanks for the insights and discussions!

To maintain ABI compatibility, I've renamed it `make_with_node_interfaces` rather than overload `make`. In the current state of this PR, lifecycle nodes will be supported for Humble users as well (if they wish to build off `rmf_ros2` `main`), just that we will have to be content with calling this slightly longwinded `make_with_node_interfaces` function,
```bash
std::shared_ptr<Auctioneer> Auctioneer::make_with_node_interfaces(
  const rclcpp::node_interfaces::NodeBaseInterface::SharedPtr node_base_interface,
  const rclcpp::node_interfaces::NodeClockInterface::SharedPtr node_clock_interface,
  const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr node_logging_interface,
  const rclcpp::node_interfaces::NodeTimersInterface::SharedPtr node_timers_interface,
  const rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr node_topics_interface,
  const rclcpp::node_interfaces::NodeParametersInterface::SharedPtr node_parameters_interface,
  BiddingResultCallback result_callback,
  ConstEvaluatorPtr evaluator)
```
rather than the previously considered and more concise function call
```
using AuctioneerNodeInterfaces = rclcpp::node_interfaces::NodeInterfaces<
    rclcpp::node_interfaces::NodeBaseInterface,
    rclcpp::node_interfaces::NodeClockInterface,
    rclcpp::node_interfaces::NodeLoggingInterface,
    rclcpp::node_interfaces::NodeTimersInterface,
    rclcpp::node_interfaces::NodeTopicsInterface,
    rclcpp::node_interfaces::NodeParametersInterface>;

std::shared_ptr<Auctioneer> make(
    AuctioneerNodeInterfaces node_interfaces,
    BiddingResultCallback result_callback,
    ConstEvaluatorPtr evaluator);
```

The intention would be to cherry-pick and backport this Jazzy, since it will be ABI compatible and does not introduce new features or behaviors. Let me know how if this is agreeable.

rmf_task_ros2/src/rmf_task_ros2/bidding/Auctioneer.cpp
```diff
@@ -34,22 +54,33 @@ Auctioneer::Implementation::Implementation(
...
+    node_topics_interface,
+    rmf_task_ros2::BidResponseTopicName,
+    dispatch_qos,
     [&](const BidResponseMsg::UniquePtr msg)
     {
       this->receive_response(*msg);
     });
 
-  timer = node->create_wall_timer(std::chrono::milliseconds(200), [&]()
+  timer = rclcpp::create_timer(
```
luca-della-vedova, line 78: It seems this changed the function from `create_wall_timer` to `create_timer`, there is a [rclcpp::create_wall_timer](https://github.com/ros2/rclcpp/blob/db8d917a12d640b832cf2ee773e338b98b69e191/rclcpp/include/rclcpp/create_timer.hpp#L191), any reason for this change?

aaronchongth: Thanks for catching that, I intended to "fix" the behavior, since a generic timer adheres to sim time as well, so that might be what we actually want. But forgot to revert this to not change any behavior. I'll switch it back to wall timer

aaronchongth: https://github.com/open-rmf/rmf_ros2/pull/420/commits/e80451e22e8457f8707d9e82c218deb0fbb482bf

luca-della-vedova (approved):
LGTM thanks for iterating!

Issue #421, opened 2025-05-07
[Bug]: Incorrect Orientation Goal Check (Angle Wrap-around and Tolerance)


alex-roba:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

24.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

main

### ROS distribution

Rolling

### ROS installation type

Source build

### Description of the bug

The code checks if a robot's current orientation (l.orientation()) is significantly different from a target orientation derived from a waypoint (wp.position()[2]). This check is performed using the condition std::abs(wp.position()[2] - l.orientation()) * 180.0 / M_PI > 1e-2. 

https://github.com/open-rmf/rmf_ros2/blob/160176cb818b78559fafd8a8ad4af453ba6e4382/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyFullControl.cpp#L1071

However, this approach presents two main problems that can lead to incorrect behavior:

First, the tolerance 1e-2 degrees (0.01 degrees) is extremely small. If the robot's actual orientation is, for instance, 179.7 degrees and the target is 180.0 degrees, the raw difference is 0.3 degrees. Since 0.3 is greater than 0.01, the condition is met, signaling that the orientation goal is not aligned. If 179.7 degrees is already within the robot's control or sensing precision relative to 180.0 degrees, subsequent commands might be sent to correct this tiny difference, but the robot may not actually move because it considers itself "on target" within its own internal tolerance. This can result in the system incorrectly believing the goal isn't met and potentially entering a loop or becoming stuck.

Second, and critically, the calculation std::abs(wp.position()[2] - l.orientation()) does not account for the inherent wrap-around nature of angles. The difference between two angles should represent the shortest angular distance between them (e.g., the difference between 179 degrees and -179 degrees is only 2 degrees, not 358 degrees). Using the absolute value of the raw difference means that if the target angle is just over the ±180 ∘
  boundary and the current angle is just under (or vice-versa), the calculated difference will be close to 360 degrees. This large, incorrect difference will always exceed the tiny tolerance, causing the condition to wrongly indicate that the orientation is far from the target when it is actually very close.

### Steps to reproduce the bug

launch

mxgrey:
[...]

PR #422, opened 2025-05-07
Fix orientation mismatch when !skip_rotation_commands

mxgrey:
This is an attempt to address https://github.com/open-rmf/rmf_ros2/issues/421

A problem was introduced by https://github.com/open-rmf/rmf_ros2/pull/398 that can cause an infinite loop of replanning. The gist of the problem is that when the user has set `!skip_rotation_commands`, the EasyFullControl implementation of `follow_new_path` will be overly strict about the initial conditions of the robot. It will test the robot's current orientation against the orientations expected at discrete points along the planned path. If the robot does not align with one of the orientations then the plan gets rejected and a replan is requested. This can happen infinitely many times.

The fix proposed by this PR is to remove that criteria and instead, when `skip_rotation_commands` is false, inject a command at the front of the queue to ensure the robot is facing the correct direction before moving towards the first target.

@alex-roba I'm having a hard time recreating the conditions to verify that this fix is working. If you can try this out and let me know if it fixes the problems you've encountered, I'd appreciate it. I plan on moving forward with a release into Kilted tomorrow, but I'll cut another release into Kilted at the earliest opportunity after this PR is merged once you let me know that it's working.

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -1,0 +1,16 @@
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+name: Ticket opened
+jobs:
+  assign:
+    name: Add ticket to inbox
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add ticket to inbox
+        uses: actions/add-to-project@v1.0.2
+        with:
+          project-url: https://github.com/orgs/open-rmf/projects/10
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
--- a/rmf_charging_schedule/CHANGELOG.rst
+++ b/rmf_charging_schedule/CHANGELOG.rst
@@ -2,6 +2,16 @@
 Changelog for package rmf_charging_schedule
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2.11.0 (2025-06-09)
+-------------------
+
+2.10.1 (2025-05-13)
+-------------------
+* Remove icecream as a dependency entirely (`#432 <https://github.com/open-rmf/rmf_ros2/issues/432>`_)
+
+2.10.0 (2025-05-09)
+-------------------
+
 2.9.0 (2024-11-27)
 ------------------
 * Fixes warning given by `colcon_ros` on Jazzy (`#377 <https://github.com/open-rmf/rmf_ros2/issues/377>`_)
--- a/rmf_charging_schedule/package.xml
+++ b/rmf_charging_schedule/package.xml
@@ -2,14 +2,13 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>rmf_charging_schedule</name>
-  <version>2.9.0</version>
+  <version>2.11.0</version>
   <description>Node for a fixed 24-hour rotating charger usage schedule</description>
   <maintainer email="mxgrey@intrinsic.ai">Grey</maintainer>
   <license>Apache License 2.0</license>
 
   <depend>rclpy</depend>
   <depend>rmf_fleet_msgs</depend>
-  <depend>python3-icecream</depend>
 
   <export>
     <build_type>ament_python</build_type>
--- a/rmf_charging_schedule/rmf_charging_schedule/main.py
+++ b/rmf_charging_schedule/rmf_charging_schedule/main.py
@@ -7,8 +7,6 @@ import bisect
 from functools import total_ordering
 from zoneinfo import ZoneInfo
 
-from icecream import ic
-
 import rclpy
 from rclpy.node import Node, Publisher
 from rclpy.qos import (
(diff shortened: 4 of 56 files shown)


alex-roba:
@mxgrey During testing, I've identified two areas for consideration. Firstly, I've observed what appears to be redundancy in navigation commands, with the robot being commanded to navigate to its current location before each path following request. To improve efficiency, I suggest implementing a check with a heading tolerance; if the robot's current angle doesn't match the required angle for the upcoming path command, a dedicated pre-orientation command could be issued first otherwise no need just go with the path commands. Secondly, while testing this pull request, I consistently encountered a warning message: [fleet_adapter-16] [WARN] ... [tinyRobot_fleet_adapter]: Relocalizing robot [tinyRobot/tinyRobot2] from map [] to [L1] because of a mismatch while starting to follow a new path).

mxgrey:
> To improve efficiency, I suggest implementing a check with a heading tolerance; if the robot's current angle doesn't match the required angle for the upcoming path command, a dedicated pre-orientation command could be issued first otherwise no need just go with the path commands

Just to be clear, the core complaint here is that we're issuing commands that aren't entirely necessary, right?

In general, I'm reluctant to add global tolerance parameters. I think we have too many already (e.g. all the different "merge waypoint" and "merge lane" parameters) that users aren't aware of or don't understand the meaning of. In particular I'm concerned that tolerances should not be global. There may be certain regions or certain situations in which tolerances should be higher or lower, and that can't be reflected by a global parameter. It also creates an uncomfortable situation of deciding on an arbitrary default.

For now I'd rather have the decision about whether a goal is within tolerance be made by the integration side of the adapter. We do have baked-in tolerances in some parts for the sake of avoiding mathematical singularities, but those tend to be reasonably universal values.

That being said, the architecture of the next generation fleet adapters will be based heavily on workflows. That will allow us to provide various command filter nodes that users can choose to insert into their command workflow. We could provide a workflow node that just uses a fixed tolerance, and we could also provide workflow nodes that set the tolerance based on location or task. Users could also define their own filtering nodes and insert them.

> [fleet_adapter-16] [WARN] ... [tinyRobot_fleet_adapter]: Relocalizing robot [tinyRobot/tinyRobot2] from map [] to [L1] because of a mismatch while starting to follow a new path).

Thanks for catching this. It was caused by the perils of move semantics in C++. It should be fixed by this commit: [519389f](https://github.com/open-rmf/rmf_ros2/pull/422/commits/519389f4eaacb9faae491ebaeb4610cd5484f0c7)

alex-roba:
@mxgrey okay for now at least need to fix the issue caused by the heading angle calculation. I've commented that section pls check.

mxgrey:
> I've commented that section pls check.

Can you provide a link to your comment? I'm not seeing any inline comments in the `Files changed` tab.

alex-roba:
@mxgrey I could not get the link i here is a screen shoot 
image

mxgrey:
> I could not get the link 

For reference, the `Pending` label on your comment means it's waiting for you to submit a full review before it will publish the comment. When you wrote it, you must have chosen `Start a review` when you meant to select `Add single comment`.

The rationale behind the current approach is the first command in the queue **should** be the first position along the path that the robot needs to be commanded to move towards, i.e. we first find where the robot currently is along the commanded path and then we cut off the portion of the path up to and including that point.

So if the first command in the queue should be where the robot needs to move to next, then we should add a new command right in front of that which positions the robot in correct heading to move towards that first point.

If the behavior that you're seeing doesn't match that, then I'll need to look into what part isn't working the way I would've expected. If you can provide a screen recording of rviz that illustrates a problem that you're seeing, it may help me figure out the problem.

I do anticipate if a robot gets suddenly rerouted there could be some sloppiness since the fleet adapter may fall a bit out of sync with the current position of the robot. For the next generation I want to provide an interface that can make it easier for system integrators to connect Open-RMF to a pure pursuit controller which should eliminate that kind of problem.

alex-roba:
@mxgrey this screen recording shows the issue I'm talking about 

https://github.com/user-attachments/assets/4d4b9411-ac15-4cdf-8e09-d470cbf727f2

mxgrey:
I see, it looks like there are two factors leading to problems: 
* The new code I put in isn't taking into account reversibility
* The first command is to go to the current position instead of the next position

I'll give some thought to how to fix this.

mxgrey:
I've put in a fix to account for reversibility: [5eaa649](https://github.com/open-rmf/rmf_ros2/pull/422/commits/5eaa649afa8ca301c01b018a09d9c504fc09875c). I did a before/after test and confirmed that I could reproduce the unnecessary rotation issue, and then confirmed that the new commit has fixed it.

In the process of investigating that, I also discovered a flaw in the slotcar plugin, and opened a PR to fix it: https://github.com/open-rmf/rmf_simulation/pull/149

@alex-roba let me know if you come across any more issues. We'll probably merge this soon since these are valuable fixes, but we can follow up with another PR if you discover any other problems.

luca-della-vedova (approved):

xiyuoh (approved):

Yadunund (approved):
LGTM

Yadunund:
Would it be possible to backport this fix to `jazzy` and bloom a new release?

Issue #425, opened 2025-05-08
[Bug]: Final Orientation in not respected in the Office world


alex-roba:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 24.04

### Open-RMF installation type

Source build

### Open-RMF version or commit hash

main

### ROS distribution

Jazzy

### ROS installation type

Binaries

### Description of the bug

the final Orientation in not respected in the Office world 

### Steps to reproduce the bug

1. source ~/rmf_demos_ws/install/setup.bash
2. ros2 launch rmf_demos_gz office.launch.xml
3. ros2 run rmf_demos_tasks  dispatch_go_to_place -p lounge -o 105 --use_sim_time

mxgrey:
Thanks for reporting this, it seems we didn't test out orientation constraints for goals before merging #325 . I've opened https://github.com/open-rmf/rmf_ros2/pull/426 to fix it.

PR #426, opened 2025-05-08
Fix goal orientation constraint

mxgrey:
This fixes a subtle bug reported by #425 introduced in #325

We were accidentally constructing a new goal using only the waypoint information of the goal that the user originally passed in. The waypoint value was being implicitly converted into a complete goal description by the compiler.


mxgrey:
I've merged this PR to make sure it gets into the kilted release.

PR #427, opened 2025-05-08
Add a timer to log the size of planner caches

mxgrey:
This builds on [#117](https://github.com/open-rmf/rmf_traffic/pull/117) by adding a timer to the fleet adapter to print out the memory footprint of the traffic planner once per minute. We could consider reducing that frequency to once per five minutes or anything else that anyone would suggest.


mxgrey:
This PR now additionally depends on https://github.com/open-rmf/rmf_traffic/pull/118

PR #444, opened 2025-06-16
Remove pybind11 vendored dependencies

luca-della-vedova:
Ref https://github.com/open-rmf/pybind11_json_vendor/issues/16, I tested locally on a noble / kilted setup and demos still work, checking CI first

### DIFF
--- a/rmf_fleet_adapter_python/CMakeLists.txt
+++ b/rmf_fleet_adapter_python/CMakeLists.txt
@@ -30,10 +30,8 @@ include_directories(
   ${rmf_fleet_adapter_INCLUDE_DIRS}
 )
 
-find_package(pybind11_vendor REQUIRED)
 find_package(pybind11 REQUIRED)
 find_package(nlohmann_json REQUIRED)
-find_package(pybind11_json_vendor REQUIRED)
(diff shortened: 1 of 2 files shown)


luca-della-vedova:
I tried it on a `humble` VM with `main` for the other core RMF repos, with an office world sending tasks and still seems to work as well

luca-della-vedova:
After discussing on the PMC I checked that I was testing the right code paths since it seems this might only be used when doing perform_action.

I tested this in the hotel world by sending a cleaning task which, to my understanding, should use perform action and specifically get the cleaning zone from the description json [here](https://github.com/open-rmf/rmf_demos/blob/febbc659e3b33866349c20ccdce461fa767ab7cf/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py#L277-L279).
It seems to work ok but happy for anyone else to have another look

aaronchongth (approved):
Tried it on Humble 22.04, Kilted 24.04, things are working well, `rosdep` also installs the released official binaries as expected. I've also made sure to not build the vendored package by removing it from https://github.com/open-rmf/rmf/blob/main/rmf.repos (which will probably need to be done too)

PR #449, opened 2025-07-02
Fix a race condition that can cause an endless loop when waiting for a lift

mxgrey:
While doing some simulation testing, I encountered a race condition which is difficult to reproduce but which can cause a robot to keep waiting for a lift even though the lift has arrived.

The race condition (which in some ways is more related to tolerances than races) happens when a lift arrives for a robot that was waiting for it. When it takes some time for the lift to arrive, the traffic planner will replan the path for the robot that was waiting, so that it's up to date with the latest traffic activity.

In some cases, if the robot isn't positioned well in the navigation graph, the plan executor might be tricked into restarting the whole process of waiting for the lift. When that happens once, it will likely keep happening indefinitely until the robot's position gets manually adjusted, creating a potentially infinite wait.

This PR fixes the problem by not allowing the lift waiting behavior to be restarted if we've gone back to waiting for the same lift as previously.

To validate this PR you can try sending robots up to L2 and L3 in the hotel world. I don't have a reliable way to reproduce it, but what usually works is to issue the commands

```bash
ros2 run rmf_demos_tasks dispatch_go_to_place --use_sim_time -p L2_master_suite
ros2 run rmf_demos_tasks dispatch_go_to_place --use_sim_time -p L3_master_suite
```

and then sometimes the second robot (`tinyBot_1`) will get stuck waiting infinitely. That should never be able to happen on this PR branch.

### DIFF
--- a/rmf_fleet_adapter_python/CMakeLists.txt
+++ b/rmf_fleet_adapter_python/CMakeLists.txt
@@ -30,10 +30,8 @@ include_directories(
   ${rmf_fleet_adapter_INCLUDE_DIRS}
 )
 
-find_package(pybind11_vendor REQUIRED)
 find_package(pybind11 REQUIRED)
 find_package(nlohmann_json REQUIRED)
-find_package(pybind11_json_vendor REQUIRED)
 find_package(pybind11_json REQUIRED)
 
 pybind11_add_module(rmf_adapter SHARED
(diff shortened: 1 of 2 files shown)


mxgrey:
> I also tried about 10 runs of trying to reproduce the issue with the recommended tasks and I couldn't on main

The failures are flaky and I don't have a specific explanation for why it happens when it does (it has something to do with how exactly the simulation plays out, not exactly a race condition inside of `rmf_fleet_adapter` itself), nor for why it was never witnessed before, I only know
1. The code path that's fixed in this PR is what was being executed when it got stuck in the loop. 
2. The fix that's been applied is not detrimental in any situation, so there's no risk of introducing it.

PR #450, opened 2025-07-16
Remove deprecated Pose2D message

luca-della-vedova:
## Bug fix

Together with https://github.com/open-rmf/rmf_internal_msgs/pull/88, closes https://github.com/open-rmf/rmf_internal_msgs/issues/87

### Fixed bug

Migrate to the new internal `Pose2D` message.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/rmf_task_ros2/CMakeLists.txt
+++ b/rmf_task_ros2/CMakeLists.txt
@@ -15,6 +15,7 @@ include(GNUInstallDirs)
 
 find_package(ament_cmake REQUIRED)
 find_package(backward_ros REQUIRED)
+find_package(std_msgs REQUIRED)
 find_package(rmf_api_msgs REQUIRED)
 find_package(rmf_traffic REQUIRED)
 find_package(rmf_traffic_ros2 REQUIRED)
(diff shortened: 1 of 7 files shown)


rmf_task_ros2/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ include(GNUInstallDirs)
 
 find_package(ament_cmake REQUIRED)
 find_package(backward_ros REQUIRED)
+find_package(std_msgs REQUIRED)
```
luca-della-vedova, line 18: This was an interesting change, cleaning up the `geometry_msgs` dependencies where they were not needed upstream removed a transitive dependency on `std_msgs` for this package that made it fail compiling because [this std_msgs header](https://github.com/open-rmf/rmf_ros2/blob/8dee3afeb9513be711b8f1c2dea9338b57bbdeae/rmf_task_ros2/src/rmf_task_ros2/Dispatcher.cpp#L49) was not found.
Since the package needs std_msgs it should include it in its package.xml and CMakeLists file explicitly rather than rely on transitive dependencies

mxgrey (approved):

PR #451, opened 2025-07-18
Adds a check for numpy version

arjo129:
### Fixed bug

This PR adds a simple check to check the version of numpy loaded by the python interpretter is the same as the version of numpy loaded by the user. This should prevent ABI incompatibilities and undefined behaviour which may arise from the user installing a newer version of numpy than what we build against.

### GenAI Use

- [ ] I did not use GenAI

Generated-by:


mxgrey:
This is a great idea with the number of segfault reports we've been seeing which ultimately turn out to be caused by inconsistent numpy versions.

mxgrey (approved):
I just left one suggestion but it's such a niche consideration that I'd be fine merging this PR either way.

mxgrey (approved):

Issue #456, opened 2025-07-25
Concerns with BroadcastServer


sauk2:
## Summary

This issue details two distinct concerns related to the `rmf_websocket::BroadcastServer` - unsafe shutdown error and a request for improved logging.

### Problem: Shutdown error (`Bad file descriptor`)

`BroadcastServer` shutdown causes the following errors,

```
[info] Error getting remote endpoint: system:9 (Bad file descriptor)
[info] asio async_shutdown error: system:9 (Bad file descriptor)
[info] handle_accept error: Operation canceled
[info] Stopping acceptance of new connections because the underlying transport is no longer listening.
```

This seems to indicate a race condition or improper cleanup of the ASIO resources. It always occurs immediately after `_data->echo_server.stop_listening()` is called within the `stop()` method

https://github.com/open-rmf/rmf_ros2/blob/a4a984a157b0a4a5323f77a37748fa06048f324e/rmf_websocket/src/rmf_websocket/BroadcastServer.cpp#L88-L99

#### Proposed fix
Ensure all server-level shutdown commands are posted to and executed on the `io_service`'s thread (addressed in #457).

### Feature Request: Replace `stdout` logging

Currently `BroadcastServer` uses `std::cout` for logging output which streams directly to standard output. I would like to propose replacing direct `std::cout` calls with ROS 2 logging. This would involve passing a node by modifying the `BroadcastServer`'s constructor.

#### API/ABI compatibility concerns

This would likely constitute an API break for `rmf_websocket` as constructor for `BroadcastServer` would change. It might also have ABI implications that I am not aware of. An initial search of the all Open-RMF repos indicates that this class is only ever used in the following tests

- https://github.com/open-rmf/rmf_ros2/blob/a4a984a157b0a4a5323f77a37748fa06048f324e/rmf_fleet_adapter/test/tasks/test_Loop.cpp#L169
- https://github.com/open-rmf/rmf_ros2/blob/a4a984a157b0a4a5323f77a37748fa06048f324e/rmf_fleet_adapter/test/tasks/test_Delivery.cpp#L374

So the question to maintainers, is this break acceptable so that we can stream logs to ROS 2 logging instead of `stdout`?

aaronchongth:
Thanks for putting this issue ticket together. I haven't looked into the fixes yet, however regarding migrating logging to ROS 2 logging, I think that's a great idea.

A new builder function can be introduced, alongside the existing `make` function.

```cpp
[...]

PR #458, opened 2025-07-31
Added RCL logging for BroadcastServer

sauk2:
### Implemented feature

Related to #456. Based on @aaronchongth's [comment](https://github.com/open-rmf/rmf_ros2/issues/456#issuecomment-3126276261) I have added a new builder function that takes in a node as an argument.

A new builder function `make_with_logger` takes in an argument of type `const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr` which is then stored in `Implementation`. The validity of this node pointer is checked each time before printing a log.

The usage of this function can be seen in [`test_Delivery.cpp`](https://github.com/sauk2/rmf_ros2/blob/b7f865a50da3c98cd0c1f39584ef80a6816bad74/rmf_fleet_adapter/test/tasks/test_Delivery.cpp#L375) of `rmf_fleet_adapter`

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/rmf_websocket/include/rmf_websocket/BroadcastServer.hpp
+++ b/rmf_websocket/include/rmf_websocket/BroadcastServer.hpp
@@ -45,7 +45,8 @@ public:
 
   using ApiMessageCallback = std::function<void(const nlohmann::json&)>;
 
-  /// Add a callback to convert from a Description into an active Task.
+  /// Create a wrapper around a websocket server for receiving states and logs for
+  /// fleets, robots and tasks
   ///
   /// \param[in] port
   ///   server url port number
@@ -61,7 +62,8 @@ public:
     ApiMessageCallback callback,
     std::optional<ApiMsgType> msg_selection = std::nullopt);
 
-  /// Add a callback to convert from a Description into an active Task.
+  /// Create a wrapper around a websocket server for receiving states and logs for
+  /// fleets, robots and tasks
   ///
   /// \param[in] port
   ///   server url port number
@@ -69,17 +71,19 @@ public:
   /// \param[in] callback
   ///   callback function when the message is received
   ///
+  /// \param[in] node_logging_interface
+  ///   node interface for logging. Default as nullptr which will result
+  ///   in errors and additional debug information not being logged
+  ///
   /// \param[in] msg_selection
   ///   selected msg type to listen. Will listen to all msg if nullopt
   ///
-  /// \param[in] node_logging_interface
-  ///   node for logging
   static std::shared_ptr<BroadcastServer> make_with_logger(
     const int port,
     ApiMessageCallback callback,
-    std::optional<ApiMsgType> msg_selection = std::nullopt,
     const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
-      node_logging_interface = nullptr);
+      node_logging_interface,
+    std::optional<ApiMsgType> msg_selection = std::nullopt);
 
   /// Start Server
   void start();
--- a/rmf_websocket/include/rmf_websocket/BroadcastServer.hpp
+++ b/rmf_websocket/include/rmf_websocket/BroadcastServer.hpp
@@ -23,6 +23,8 @@
 #include <optional>
 #include <rmf_utils/impl_ptr.hpp>
 
+#include <rclcpp/rclcpp.hpp>
+
 namespace rmf_websocket {
 
 //==============================================================================
@@ -58,6 +60,26 @@ public:
     const int port,
     ApiMessageCallback callback,
     std::optional<ApiMsgType> msg_selection = std::nullopt);
+
+  /// Add a callback to convert from a Description into an active Task.
+  ///
+  /// \param[in] port
+  ///   server url port number
+  ///
+  /// \param[in] callback
+  ///   callback function when the message is received
+  ///
+  /// \param[in] msg_selection
+  ///   selected msg type to listen. Will listen to all msg if nullopt
+  ///
+  /// \param[in] node_logging_interface
+  ///   node for logging
+  static std::shared_ptr<BroadcastServer> make_with_logger(
+    const int port,
+    ApiMessageCallback callback,
+    std::optional<ApiMsgType> msg_selection = std::nullopt,
+    const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
+      node_logging_interface = nullptr);
 
   /// Start Server
   void start();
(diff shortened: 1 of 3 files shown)


rmf_websocket/include/rmf_websocket/BroadcastServer.hpp
```diff
@@ -59,6 +61,26 @@ class BroadcastServer
...
+  ///   selected msg type to listen. Will listen to all msg if nullopt
+  ///
+  /// \param[in] node_logging_interface
+  ///   node for logging
+  static std::shared_ptr<BroadcastServer> make_with_logger(
+    const int port,
+    ApiMessageCallback callback,
+    std::optional<ApiMsgType> msg_selection = std::nullopt,
+    const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
+      node_logging_interface = nullptr);
```
aaronchongth, line 82:
```suggestion
  static std::shared_ptr<BroadcastServer> make_with_logger(
    const int port,
    ApiMessageCallback callback,
    const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
      node_logging_interface,
    std::optional<ApiMsgType> msg_selection = std::nullopt);
```

On second thought, since we are introducing a new API, we should avoid the default behavior (if optional arguments are not provided) to be the same as the original API. I think it's a better idea to ensure that this API is specific for introducing the logging interface as a non-optional argument

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -39,10 +39,19 @@ class BroadcastServer::Implementation
   Implementation(
     const int port,
     ApiMessageCallback callback,
-    std::optional<ApiMsgType> msg_selection)
-  : _data(std::make_shared<Data>(std::move(callback),
-      std::move(msg_selection)))
+    std::optional<ApiMsgType> msg_selection,
+    const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
+      node_logging_interface = nullptr)
```
aaronchongth, line 44:
```suggestion
      node_logging_interface)
```

We can avoid using default arguments in `Implementation` constructors, by choosing what the static `make` functions pass into them. So the original `make` functions implementation can pass in a `nullptr` instead

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -87,6 +121,12 @@ class BroadcastServer::Implementation
   {
     if (_server_thread.joinable())
     {
+      if (_logger_interface)
+      {
+        RCLCPP_INFO(
+          _logger_interface->get_logger(),"Stopping BroadcastServer");
```
aaronchongth, line 127:
```suggestion
          _logger_interface->get_logger(), "Stopping BroadcastServer");
```

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -113,6 +153,7 @@ class BroadcastServer::Implementation
     return map.at(type);
   }
 
+
```
aaronchongth, line 156: nit: extra line here

aaronchongth (changes requested):
Thanks for adding this feature, it's looking good! I apologize for doubling back on my initial API suggestion with some changes, but this should help us be more specific with the behavior and avoid defaults. I left some comments

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -39,10 +39,19 @@ class BroadcastServer::Implementation
   Implementation(
     const int port,
     ApiMessageCallback callback,
-    std::optional<ApiMsgType> msg_selection)
-  : _data(std::make_shared<Data>(std::move(callback),
-      std::move(msg_selection)))
+    std::optional<ApiMsgType> msg_selection,
+    const rclcpp::node_interfaces::NodeLoggingInterface::SharedPtr
+      node_logging_interface)
+  : _data(std::make_shared<Data>(std::move(callback), std::move(msg_selection))),
```
aaronchongth, line 45:
```suggestion
  : _data(std::make_shared<Data>(std::move(callback),
    std::move(msg_selection))),
```
nit: line longer than 80 characters

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -39,10 +39,19 @@ class BroadcastServer::Implementation
...
+    _logger_interface(node_logging_interface)
   {
+    if(_logger_interface)
```
aaronchongth, line 48:
```suggestion
    if (_logger_interface)
```

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -66,17 +75,42 @@ class BroadcastServer::Implementation
     }
     catch (const websocketpp::exception& e)
     {
-      std::cout << e.what() << std::endl;
+      if(_logger_interface)
```
aaronchongth, line 78:
```suggestion
      if (_logger_interface)
```

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -66,17 +75,42 @@ class BroadcastServer::Implementation
...
+      {
+        RCLCPP_ERROR_STREAM(
+          _logger_interface->get_logger(),
+          "WebSocket exception occured: " << e.what());
+      }
     }
     catch (...)
     {
-      std::cout << "other exception" << std::endl;
+      if(_logger_interface)
```
aaronchongth, line 87:
```suggestion
      if (_logger_interface)
```

rmf_websocket/src/rmf_websocket/BroadcastServer.cpp
```diff
@@ -66,17 +75,42 @@ class BroadcastServer::Implementation
...
   }
 
   /// Start Server
   void start()
   {
+    if (_logger_interface)
+    {
+      websocketpp::lib::asio::error_code ec;
+      auto endpoint = _data->echo_server.get_local_endpoint(ec);
+      if(ec)
```
aaronchongth, line 102:
```suggestion
      if (ec)
```

aaronchongth (changes requested):
Thanks for making the changes! I only have a few lint related comments, otherwise LGTM

sauk2:
> Thanks for making the changes! I only have a few lint related comments, otherwise LGTM

Awesome! Thanks for the review :tada:

PR #459, opened 2025-08-07
Backporting BroadcastServer improvements to Jazzy

sauk2:
This PR backports the improvements to `BroadcastServer` as described in #456, #457 and #458


aaronchongth (changes requested):
Thanks for handling the backport! We generally try to backport by cherry-picking the exact commit/s, so that we can merge commits to keep history (instead of squashing). In this case, it will be cherry-picking https://github.com/open-rmf/rmf_ros2/commit/f41fae50cab9c008853632f6f0a229b60c7c6994 specifically to `jazzy`

PR #464, opened 2025-08-19
Remove potential data races

### LINKED ISSUE #462: [Bug]: fleet adapter crashed intermittently while searching for emergency pullover spot

### OS and version

Ubuntu 22.04

Source build

_No response_

f41fae50cab9c008853632f6f0a229b60c7c6994

### ROS distribution

Humble

### ROS installation type

Binaries

_No response_

_No response_

### Description of the bug

Tested using airport_terminal simulation with multiple robots performing dispatch patrol task. 
The emergency signal was triggered using `emergency_signal` topic to all fleets and it took a few attempts of toggling of the emergency signal before it crashed. 

[gdb output](https://gist.github.com/cwrx777/11f2fd6984a7c465c737f97689e602a3)

Updated [rmf_demos_fleet_adapter.launch.xml](https://github.com/open-rmf/rmf_demos/blob/humble/rmf_demos_fleet_adapter/launch/fleet_adapter.launch.xml) file with gdb as launch prefix:
```xml
  <!-- Fleet adapter -->
  <group if="$(var use_sim_time)">
    <node pkg="rmf_demos_fleet_adapter"
      exec="fleet_adapter"
      args="-c $(var config_file) -n $(var nav_graph_file) -sim"
      output="both"
      launch-prefix="xterm -e gdb -ex=r --args python3"
    >
...
```
Launched airport_terminal simulation
```bash
ros2 launch rmf_demos_gz_classic airport_terminal.launch.xml
```

Send tasks
```bash
ros2 run rmf_demos_tasks dispatch_patrol --use_sim_time -p koi_pond junction_s10 -n 5; \
ros2 run rmf_demos_tasks dispatch_patrol --use_sim_time -p junction_s10 koi_pond -n 5; \
ros2 run rmf_demos_tasks dispatch_patrol --use_sim_time -p junction_s10 koi_pond -n 5; \
ros2 run rmf_demos_tasks dispatch_patrol --use_sim_time -p deliveryRobot_n10 junction_north_east -n 5; \
ros2 run rmf_demos_tasks dispatch_patrol --use_sim_time -p junction_north_east deliveryRobot_n10 -n 5; \
ros2 run rmf_demos_tasks dispatch_patrol --use_sim_time -p junction_north_east deliveryRobot_n10 -n 5;
```
Toggle emergency signal multiple times
```
while true; do
  # Trigger emergency = true
  ros2 topic pub /emergency_signal rmf_fleet_msgs/msg/EmergencySignal "
fleet_names: []
is_emergency: true" --once
  
  sleep 10
  
  # Trigger emergency = false
  ros2 topic pub /emergency_signal rmf_fleet_msgs/msg/EmergencySignal "
fleet_names: []
is_emergency: false" --once
  
  sleep 10
done
```

### Expected behavior

Should not crash.

### Actual behavior

Fleet adapter crashed

_No response_

mxgrey:
This PR fixes #462 

Initially I thought the crashing was related to the use of loose raw pointers in `ProgressEvaluator`, so most of the changes in this PR are to replace those raw pointers with `std::shared_ptr`. While those raw pointers may have been the cause of a small fraction of the crashes, I believe most of the crashes were actually related to the `_resume` callback.

It turns out that there was a narrow race condition in which it was possible for the `_resume` callback of `Planning` to get overwritten while it was being executed. As the `std::function` is being dropped, its captured variables become invalid memory. If the `std::function` is being run **_while_** it gets dropped, those captured variables can become invalid memory **_while they are being used_**.

There was already a `_resume_mutex` being used to prevent this situation, but its usage wasn't rigorous enough. There was still a very slim window in between when the `_resume` callback gets called and when it gets locked inside the callback. If the `std::function` gets dropped inside that narrow window then we can get a segmentation fault.

I've updated the way `_resume` works so that it only gets instantiated once: the first time the planning job is started. After that there is no risk of the `std::function _resume` becoming invalid as long as the `Planning` object is still valid, and there is no way to trigger `_resume` without the `Planning` object being alive. This should guarantee all the memory remains valid.

Furthermore I've added a `_resume_scheduled` flag that tracks whether a `Planning` job has already been scheduled to resume. This makes sure we never have multiple threads trying to make progress on the same planning job at the same time.

As an extra layer of protection, I've opened https://github.com/open-rmf/rmf_traffic/pull/124 which will guarantee that access to the `Planner::Result` class is always thread-safe, in case there are other places in this codebase where a data race may be happening that we haven't caught yet.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/Planning.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/Planning.cpp
@@ -41,6 +41,12 @@ Planning::Planning(rmf_traffic::agv::Planner::Result _setup)
 //==============================================================================
 void Planning::resume()
 {
+  if (_resume_scheduled.exchange(true))
+  {
+    // This has already been scheduled to resume
+    return;
+  }
+
   _resume();
 }
 
@@ -60,25 +66,13 @@ bool Planning::active() const
 //==============================================================================
 rmf_traffic::agv::Planner::Result& Planning::progress()
 {
-  return *_current_result;
+  return _current_result.value();
 }
 
 //==============================================================================
 const rmf_traffic::agv::Planner::Result& Planning::progress() const
 {
-  return *_current_result;
-}
-
-//==============================================================================
-std::unique_lock<std::mutex> Planning::_lock_resume() const
-{
-  std::unique_lock<std::mutex> lock(_resume_mutex, std::defer_lock);
-  while (!lock.try_lock())
-  {
-    // Intentionally busy wait to obtain the mutex as fast as possible
-  }
-
-  return lock;
+  return _current_result.value();
 }
 
 } // namespace jobs
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/Planning.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/Planning.hpp
@@ -23,6 +23,8 @@
 #include <rmf_traffic/agv/RouteValidator.hpp>
 #include <rmf_traffic/schedule/Snapshot.hpp>
 
+#include <atomic>
+
 namespace rmf_fleet_adapter {
 namespace jobs {
 
@@ -58,11 +60,9 @@ public:
   const rmf_traffic::agv::Planner::Result& progress() const;
 
 private:
-  mutable std::mutex _resume_mutex;
+  std::atomic_bool _resume_scheduled = {false};
   std::function<void()> _resume;
-  rmf_utils::optional<rmf_traffic::agv::Planner::Result> _current_result;
-
-  std::unique_lock<std::mutex> _lock_resume() const;
+  std::optional<rmf_traffic::agv::Planner::Result> _current_result;
 };
 
 } // namespace jobs
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/SearchForPath.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/SearchForPath.cpp
@@ -100,27 +100,27 @@ void SearchForPath::interrupt()
 }
 
 //==============================================================================
-Planning& SearchForPath::greedy()
+const std::shared_ptr<Planning>& SearchForPath::greedy()
 {
-  return *_greedy_job;
+  return _greedy_job;
 }
 
 //==============================================================================
-const Planning& SearchForPath::greedy() const
+std::shared_ptr<const Planning> SearchForPath::greedy() const
 {
-  return *_greedy_job;
+  return _greedy_job;
 }
 
 //==============================================================================
-Planning& SearchForPath::compliant()
+const std::shared_ptr<Planning>& SearchForPath::compliant()
 {
-  return *_compliant_job;
+  return _compliant_job;
 }
 
 //==============================================================================
-const Planning& SearchForPath::compliant() const
+std::shared_ptr<const Planning> SearchForPath::compliant() const
 {
-  return *_compliant_job;
+  return _compliant_job;
 }
 
 //==============================================================================
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/SearchForPath.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/jobs/SearchForPath.hpp
@@ -19,6 +19,7 @@
 #define SRC__RMF_FLEET_ADAPTER__JOBS__SEARCHFORPATH_HPP
 
 #include "Planning.hpp"
+#include <memory>
 
 namespace rmf_fleet_adapter {
 namespace jobs {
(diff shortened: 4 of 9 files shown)


xiyuoh:
While I haven't been able to reproduce the crashes, I did come across a few occasions where this warning pops up:
```
[fleet_adapter-18] -------------------------------------------------
[fleet_adapter-18] [rmf_traffic::agv::Planner::plan] WARNING: A rare anomaly has occurred in the planner. The Route Validator has failed to recognize a specified route dependency: 11 on {3 39124 0 88}. These are the trajectories:
...
(printout of some trajectories)
...
[fleet_adapter-18] Please provide this information to the RMF developers for debugging.
[fleet_adapter-18] -------------------------------------------------
```
which could be due to the undefined behavior as well, though it does not happen 100% of the time.

After extensive testing with these fixes, the above warning has not resurfaced. However, in the latest run on `main` with Jazzy, `rmf_traffic_schedule` crashed with the following logs during an active negotiation:
```
[rmf_traffic_schedule-1] terminate called after throwing an instance of 'std::bad_array_new_length'
[rmf_traffic_schedule-1]   what():  std::bad_array_new_length
[rmf_traffic_schedule-1] Stack trace (most recent call last) in thread 326159:
[rmf_traffic_schedule-1] #13   Object "/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2", at 0xffffffffffffffff, in 
[rmf_traffic_schedule-1] #12   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f79a2e3bc3b, in 
[rmf_traffic_schedule-1] #11   Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f79a2daeaa3, in 
[rmf_traffic_schedule-1] #10   Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f79a3010db3, in 
[rmf_traffic_schedule-1] #9    Object "/home/xiyuoh/ws_osrf/rmf_jazzy_ws/install/rmf_traffic_ros2/lib/librmf_traffic_ros2.so", at 0x7f79a3d6753c, in std::thread::_State_impl<std::thread::_Invoker<std::tuple<rmf_traffic_ros2::schedule::ScheduleNode::setup_conflict_topics_and_thread()::$_5> > >::_M_run()
[rmf_traffic_schedule-1] #8    Object "/home/xiyuoh/ws_osrf/rmf_jazzy_ws/install/rmf_traffic_ros2/lib/librmf_traffic_ros2.so", at 0x7f79a3d5841c, in rmf_traffic_ros2::schedule::ScheduleNode::publish_negotiation_states()
[rmf_traffic_schedule-1] #7    Object "/home/xiyuoh/ws_osrf/rmf_jazzy_ws/install/rmf_traffic_ros2/lib/librmf_traffic_ros2.so", at 0x7f79a3e0ea1e, in void std::vector<rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> >, std::allocator<rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> > > >::_M_realloc_insert<rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> > const&>(__gnu_cxx::__normal_iterator<rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> >*, std::vector<rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> >, std::allocator<rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> > > > >, rmf_traffic_msgs::msg::NegotiationState_<std::allocator<void> > const&)
[rmf_traffic_schedule-1] #6    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f79a2fc9a6e, in 
[rmf_traffic_schedule-1] #5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f79a2fc9a54, in std::terminate()
[rmf_traffic_schedule-1] #4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f79a2fdf0d9, in 
[rmf_traffic_schedule-1] #3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.33", at 0x7f79a2fc9ff4, in 
[rmf_traffic_schedule-1] #2    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f79a2d3a8fe, in abort
[rmf_traffic_schedule-1] #1    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f79a2d5727d, in raise
[rmf_traffic_schedule-1] #0    Object "/usr/lib/x86_64-linux-gnu/libc.so.6", at 0x7f79a2db0b2c, in pthread_kill
[rmf_traffic_schedule-1] Aborted (Signal sent by tkill() 326026 1156290)
[fleet_adapter-18] [INFO] [1756202046.300295104] [deliveryRobot_command_handle]: Commanding [deliveryRobot_2] to navigate to [ 1.13533490e+01 -1.33808559e+01 -1.37049913e-03] on map [L1]: cmd_id 144
[ERROR] [rmf_traffic_schedule-1]: process has died [pid 326026, exit code -6, cmd '/home/xiyuoh/ws_osrf/rmf_jazzy_ws/install/rmf_traffic_ros2/lib/rmf_traffic_ros2/rmf_traffic_schedule --ros-args -r __node:=rmf_traffic_schedule_primary --params-file /tmp/launch_params_krg3vqwl'].
```
Longer logs available in this [gist](https://gist.github.com/xiyuoh/968c9563a2fc880aa5be68855b9e980f).

It looks like it's coming from a different part of the code, so I doubt it's introduced by this PR.

xiyuoh:
I was able to do some testing on a separate machine and the crashes are very reproducible now. I can verify that this PR fixes those crashes.

I also left the test case running overnight on both machines with the fix, and the `bad_array_new_length` error has not reoccurred.

xiyuoh (approved):

Issue #463, opened 2025-08-19
[Other issue]: `task_id` in the `fleet_state` topic and `update_handle.current_task_id`


cwrx777:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf/open-rmf-general/103).

### Description

This question concerns the behavior of `task_id` in the `fleet_state` topic and `update_handle.current_task_id`, specifically during emergency pullover scenarios and when the interrupted task is killed or canceled. There seems to be inconsistencies 

ROS2: humble
OS: Ubuntu 22.04
rmf_ros2 commit hash: `[f41fae50cab9c008853632f6f0a229b60c7c6994](https://github.com/open-rmf/rmf_ros2/commit/f41fae50cab9c008853632f6f0a229b60c7c6994)`

## Observations:

**Scenario 1: Set Emergency signal when robot is idle ( finishing request: park)** 
| Robot State                    | Event / Action                        | `fleet_state.task_id` | `update_handle.current_task_id`       |
| ------------------------------ | ------------------------------------- | --------------------- | ------------------------------------- |
| Idle                           | Idle task set (e.g., ParkRobot)       | *(empty)*             | ParkRobotXXX                          |
| Idle                           | Set Emergency signal                  | *(empty)*             | emergency\_pullover-0                 |
| Performing emergency pullover  | Clear Emergency signal                | *(empty)*             | emergency\_pullover-0                 |
| Idle                           | Set Emergency signal                  | *(empty)*             | emergency\_pullover-1                 |
| Performing emergency pullover  | Clear Emergency signal                | *(empty)*             | emergency\_pullover-1                 |

**Scenario 2: Set Emergency signal when robot is idle (finishing request: nothing)**
| Robot State                    | Event / Action                        | `fleet_state.task_id` | `update_handle.current_task_id`       |
| ------------------------------ | ------------------------------------- | --------------------- | ------------------------------------- |
| Idle                           | Idle task set to nothing              | *(empty)*             | *(empty)*                             |
| Idle                           | Set Emergency signal                  | *(empty)*             | emergency\_pullover-0                 |
| Performing emergency pullover  | Clear Emergency signal                | *(empty)*             | emergency\_pullover-0                 |
| Idle                           | Set Emergency signal                  | *(empty)*             | emergency\_pullover-1                 |
[...]

PR #465, opened 2025-08-20
handling of task_api_requests on active task during emergency pullover

cwrx777:
Related to open-rmf/rmf_ros2#463

This PR implement the following enhancements:
1. `task_api_requests` on `active_task` during emergency pullover will be ignored.
2. After emergency signal is cleared, robot context `current_task_id` will be reset. 
3. After emergency pullover is cancelled, it will stop the robot. What the robot do if there is no active task is per finishing_request.

### GenAI Use

- [ x] I did not use GenAI

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -1737,12 +1737,6 @@ void TaskManager::_begin_next_task()
 //==============================================================================
 void TaskManager::_begin_pullover()
 {
-
-  if (_emergency_pullover && !_emergency_pullover.is_finished())
-  {
-    return;
-  }
-
   _finished_waiting = false;
   auto task_id = "emergency_pullover." + _context->name() + "."
     + _context->group() + "-"
@@ -1972,7 +1966,12 @@ void TaskManager::_resume_from_emergency()
           {"emergency finished"},
           self->_context->now());
         self->_emergency_pullover_interrupt_token = std::nullopt;
-        
+
+        RCLCPP_INFO(
+          self->_context->node()->get_logger(),
+          "Resume execution of task [%s] for [%s] after emergency pullover",
+          self->_active_task.id().c_str(),
+          self->_context->requester_id().c_str());
         self->_context->current_task_id(self->_active_task.id());
       }
       else
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -1527,6 +1527,11 @@ bool TaskManager::cancel_task(
 {
   if (_active_task && _active_task.id() == task_id)
   {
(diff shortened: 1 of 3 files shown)


mxgrey:
@cwrx777 Sorry, I'm catching up on my backlog.

The code looks good. I'll start manually testing and verifying the behavior changes soon.

One small concern I have is that it might be confusing to system integrators that queued tasks can be cancelled during emergencies while active tasks cannot. I'm not suggesting that this is the wrong approach, but it's something to think about. An alternative to blocking the various cancellation and skip commands would be to queue them up and apply them as soon as the emergency is over.

Another thing to consider is that we may want to allow active tasks to be killed during emergencies even if we do not allow them to be cancelled. Killing a task differs from cancelling in that a killed task is supposed to immediately halt anything it's doing no matter what is happening, whereas cancelling allows a safe wind down (potentially even including a cancellation phase). Theoretically it shouldn't disrupt anything to kill an active task during an emergency.

cwrx777:
> One small concern I have is that it might be confusing to system integrators that queued tasks can be cancelled during emergencies while active tasks cannot. I'm not suggesting that this is the wrong approach, but it's something to think about. An alternative to blocking the various cancellation and skip commands would be to queue them up and apply them as soon as the emergency is over.

In the PR, I implemented to ignore cancelation of active task because of the current behavior where [_begin_pullover](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp#L1725) was called again when the `cancel_task_request` for the active task is handled. I believe this is called from the [interruption callback. ](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp#L130). 
I also notice that after the emergency signal is cleared, even tough the task is supposedly already cancelled, the robot will still resume the task by moving to the target destination as described in Scenario 4 [here](https://github.com/open-rmf/rmf_ros2/issues/463#issue-3332780373). This also happened if the task is killed.

> Another thing to consider is that we may want to allow active tasks to be killed during emergencies even if we do not allow them to be cancelled. Killing a task differs from cancelling in that a killed task is supposed to immediately halt anything it's doing no matter what is happening, whereas cancelling allows a safe wind down (potentially even including a cancellation phase). Theoretically it shouldn't disrupt anything to kill an active task during an emergency.

In order to be less confusing for system integrators to define and document a clear/consistent (and configurable?) behavior when handling `task_api_request` during emergency situation (and after) and let the SI decide whether to call or not.   😃

xiyuoh (approved):
Thanks for your contribution @cwrx777 ! I've been testing this PR out with various api requests and I believe it is working as intended. It's also great to see the task ID being properly reset after ending an emergency pullover.

Intuitively I was expecting the robot's task to be successfully canceled even if the cancellation was submitted during emergency pullover. However after some offline discussion with @mxgrey it seems to require further implementation to get task cancellation and interruptions to behave as desired, and also for us to figure out what this desired behavior would look like.

We can discuss it further during PMC or community meetings, but for now this PR LGTM!

PR #466, opened 2025-08-22
Feature/Support Idle Robot Preferred in Task Planning

kjchee:
### Implemented feature

This PR is to support the new feature proposed at PR in rmf_task open-rmf/rmf_task#129

A new parameter in fleet_config.yaml (**task_assignment_strategy**) is read and passed to EasyFullControl and to FleetUpdateHandle and finally to TaskPlanner.

Also, there is new functions added in `rmf_task::State`:
```
bool is_idle() const;
State& idle(bool is_idle);
```
purpose is to reflect whether the robot is in idle state, and it is set in TaskManager.cpp `TaskManager::expected_finish_state()`. Then this idle state will be stored in initial_states and passed to TaskPlanner.

### GenAI Use

- [x] I did not use GenAI

Generated-by:

### DIFF
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/EasyFullControl.hpp
@@ -848,6 +848,13 @@ public:
   /// \sa strict_lanes
   std::unordered_set<std::size_t>& change_strict_lanes();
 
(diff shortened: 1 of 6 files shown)


mxgrey (approved):
I think this is good to go. I've opened #501 as a reminder to add python bindings for these new settings, but I don't think that needs to block this PR.

PR #468, opened 2025-08-29
Fix indefinite growth of active_dispatch_states

kjchee:
### Fixed bug

Finished-dispatched tasks remained in `active_dispatch_states` after being added to `finished_dispatch_states`, causing unbounded growth. This PR makes `move_to_finished()` erase the task from `active_dispatch_states` after moving it to `finished_dispatch_states`.

Reproduce the bug:
1. send a rmf task using `dispatch_task_request`, eg. using dispatch_patrol.py with `payload['type'] = 'dispatch_task_request'`
2. After the task is awarded to any robot, echo the topic /dispatch_states.
3. Can see those dispatched tasks still remain in `active` field

### Fix applied

Task is erased from `active_dispatch_states` in `move_to_finished()` .

### GenAI Use

- [x] I did not use GenAI

Generated-by:


mxgrey (approved):
Thanks for catching this!

PR #470, opened 2025-09-01
Fix compliant path planning interrupter to prevent unnecessary CPU usage

kjchee:
### Fixed bug

The compliant planning job never received an interrupter, meaning it would not stop when the deadline or interrupt_flag was triggered. This could allow compliant planning to continue running in the background until it eventually finished, even after an interrupt was requested.

In SearchForPath.cpp, the interrupter was mistakenly applied twice to the greedy planner options, leaving the compliant planner without one.

### Fix applied
Correct it back to apply interrupter to Compliant Planning. 

### GenAI Use

- [x] I did not use GenAI

Generated-by:


luca-della-vedova (approved):

mxgrey (approved):
Thanks for catching this subtle copy/paste bug!

PR #473, opened 2025-09-12
Fix mutex group erasure bug in _retain_mutex_groups

piliwilliam0306:
### Fixed bug

The `_retain_mutex_groups` function had a bug where it was always erasing entries from the `_locked_mutex_groups` map, even when processing the `_requesting_mutex_groups` map. 

### Fix applied

In `_retain_mutex_groups`, change the erasure from `_locked_mutex_groups` to the passed `groups` map

### GenAI Use

- [x] I did not use GenAI


mxgrey (approved):
Thanks for catching this. It looks like this is a bug that was introduced during refactoring.

mxgrey (approved):

PR #474, opened 2025-09-21
Fix tool_power_drain bug

piliwilliam0306:
### Fixed bug

`tool_power_drain` is set to `ambient_system["power"]`
```
tool_power_drain = ambient_system["power"].as<double>();
```
### Fix applied

should be:
```
tool_power_drain = tool_system["power"].as<double>();
```
### GenAI Use

- [x] I did not use GenAI

Generated-by:


xiyuoh (approved):
Great catch, thanks for the contribution!

PR #484, opened 2025-10-09
Makes sure that the reservation system always assigns nearest goal during Emergency Pullover

### LINKED ISSUE #481: [Other issue]: Parking spot selection during emergency is not the cheapest if the robot has an existing reservation

### Description

rmf_ros2 branch: [main](https://github.com/open-rmf/rmf_ros2/commit/88581db47e24ac1f11db1d19008a6ae710907fea)
rmf_demos: [humble](https://github.com/open-rmf/rmf_demos/tree/humble)

Steps to replicate:
1. Modify r[mf_demos_fleet_adapte](https://github.com/open-rmf/rmf_demos/blob/humble/rmf_demos_fleet_adapter/rmf_demos_fleet_adapter/fleet_adapter.py#L166)r to enable parking spot reservation
```python
cmd_handle.update_handle.use_parking_reservation_system(True)
cmd_handle.update_handle.enable_responsive_wait(False)
```
2. Modify `office.building.yaml` to set `is_parking_spot` of `hardware_2` and add `speed_limit` to the lane:
Basically this is to add cost to reach the parking spot.
Image

3. Build and run simulation
4. Send `tinyRobot1` to `hardware_2`
```bash
ros2 run rmf_demos_tasks dispatch_patrol -F tinyRobot -R tinyRobot1 -p hardware_2  --use_sim_time
```
5. Trigger emergency
```bash
ros2 topic pub /fire_alarm_trigger std_msgs/msg/Bool "
data: true" --once
```
Result:
Instead going to the nearest parking spot, the robot still proceed to `hardware_2`.

arjo129:
## Bug fix

Fixes #481.

### Fixed bug

This commit makes sure that the reservation system always assigns the nearest goal to a robot during an emergency pullover.

### Fix applied

Adds a param to skip the check to retain last reservation and forces the reservation system to reassign a goal during emergency pull overs.

Also some minor code style clean ups. The reservation system has some sloppy C++ and I should revisit and clean up the code a little.

### Test It

You can use the following test script to test the changes: https://github.com/arjo129/stress_test_rmf_chope/blob/main/test_emergency_pullover.bash

### GenAI Use

- [x] I did not use GenAI

Generated-by:

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.hpp
@@ -33,9 +33,6 @@ public:
   void replace_ticket(
     const rmf_reservation_msgs::msg::ReservationAllocation new_allocation);
 
-  /// Cancels the current reservation
-  void cancel();
-
   /// Retrieves the location name of the current reservation. Returns empty string if
   /// no location is found.
   std::string get_reserved_location() const;


mxgrey:
@cwrx777 would you be able to confirm for us whether this resolves #481? It looks to me like it should.

mxgrey:
> but i wonder if it's better if the flag is check outside of the for loop.

I agree, we could have

```cpp
if (!always_recalculate_nearest_goal) 
{
  for (std::size_t i =0; i < negotiator->_goals.size(); ++i)
  {
    if (events::wp_name(/* ... */) == context->_get_reserved_location())
    {
      /* ... */
    }
  }
}
```

Besides the minor performance benefit of entirely skipping the for-loop when the flag is false, I think this structure would make it a bit easier to follow the logic of this part of the code.

PR #490, opened 2025-11-19
Fix potential deadlocks from execution of mutex lock and release

### LINKED ISSUE #448: [Bug]: Title: Unexpected deadlock caused by Mutex

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 24.04

Source build

https://github.com/open-rmf/rmf

main-12b5980859562a11cb4358c0dc0e34fb9fcec7a6

### ROS distribution

Jazzy

### ROS installation type

Source build

_No response_

rmf_fleet_adapter

### Description of the bug

We have confirmed the occurrence of a deadlock due to an unexpected Mutex lock waiting state.
Mutex is set as shown in the following attached image for the map in the nearby area where the problem occurred.
Please check rmf_demos/rmf_demos_maps/maps/test/test.building.yaml in the attached demo package for details of the map.

Image

Rviz images of the problem are shown below.

Image

Each robot is submitting a task that should be moving downward.
At this time, strangely enough, tinyRobot/tinyRobot5 is trying to lock tinyRobot5_charger_mutex again, even though it has already left.
At this time, a deadlock occurs because tinyRobot/tinyRobot4 has already acquired tinyRobot5_charger_mutex.

```
[fleet_adapter-17] [INFO] [1750310896.033224395] [tinyRobot_fleet_adapter]: [tinyRobot/tinyRobot5] is waiting to lock mutex group [tinyRobot5_charger_mutex] but that mutex is currently held by [tinyRobot/tinyRobot4]
```

It is also strange that the yellow marker is shifted from the robot's position.
I would like to ask your advice on whether I should modify the code or the NavigationGraph.
I am also attaching the Rosbag as it appeared when the problem occurred.

[rosbag2_2025_06_19-14_24_42.zip](https://github.com/user-attachments/files/20829096/rosbag2_2025_06_19-14_24_42.zip)

1.[rmf_repos.zip](https://github.com/user-attachments/files/20829105/rmf_repos.zip) is used to build the RMF environment.
2.replace the demonstrations repository.
3.RMF launch
```
ros2 launch rmf_demos_gz test.launch.xml
```
4.Task submission
```
ros2 launch rmf_demos test_loop.launch.xml
```

### Expected behavior

Each robot runs without deadlock caused by Mutex.

### Actual behavior

Deadlock is caused by Mutex.

_No response_

mxgrey:
This PR fixes #448

I found two unrelated problems which had a risk of causing deadlocks in mutex usage:

### 1. Premature release of a mutex after replanning

The `MoveRobot` phase is set up to automatically release mutexes that are no longer needed by the remaining movement that the robot is performing. In general this works fine since a robot should only need to retain mutexes that it is currently moving through.

However, there is an edge case: If replanning gets forced by a negotiation then a new plan may involve one initial step where the robot is commanded to the midpoint of a lane. In the current implementation, this midpoint is not explicitly associated with any part of the graph, i.e. it has no "graph elements" to reference to determine what mutexes should be locked or not. This leaves the `MoveRobot` phase unaware that it needs to retain the mutex of the lane that the robot is currently on, nor the mutex of any vertex that the robot is moving towards. Without that awareness, the `MoveRobot` phase will naturally release those mutexes.

In this PR, we now detect this edge case directly, and we block the `MoveRobot` phase from doing any automatic release of mutexes when it's handling this type of midpoint. This is a very blunt force approach to solving the problem, but I don't think we'll find a better approach until we reimplement the traffic system to do geometric reasoning instead of pure graph vertex/edge reasoning.

### 2. Incorrect cleanup of `current_mutex_groups` in `ExecutePlan::make`

The logic when creating the plan execution is terribly convoluted because of the need to iterate through waypoints while only inserting events as needed. The variable `current_mutex_groups` is meant to track which mutex groups need to be locked before the next phase can be started, but its value was not getting reset to `nullopt` after being used in certain cases. This meant that the plan would involve repeatedly trying to relock mutexes that had just been correctly released.

The fix proposed [in this comment](https://github.com/open-rmf/rmf_ros2/issues/448#issuecomment-3006789630) does have the effect of forcing `current_mutex_groups` to be reset, which is why it appeared to resolve this part of the deadlocking problem. However it also has a side-effect of forcing a `LockMutexGroup` phase in between every waypoint in the plan, which is why I did not go with that proposed solution.

### Other notes

Besides fixing the above problems, I've left some debug logging in this PR with the information that helped me resolve this issue. If we notice anything else suspicious about the mutex behavior in the future, those debug outputs should help resolve it a bit faster.


mxgrey:
@xiyuoh thanks for finding this additional case. I think it's worth opening a PR for this map you created in `rmf_site_ros2` so we have this as a test case that we can refer back to.

I believe what's happening in the deadlock case that you've found is `TinyRobot_4` is within the "merge lane" distance of the `{ loc_6 -> loc_8 }` lane, making its current position on that lane an "acceptable" starting point for the planner. Geometrically the cost calculation is no worse whether the robot starts "on" that lane or "on" the `loc_8` waypoint since it's still the same point in space. The planner will end up arbitrarily choosing between those two semantically different starting points because the apparent cost is the same between them. Unfortunately the `{ loc_6 -> loc_8 }` lane has the constraint that it needs to have `mutex_3` locked, which is something that doesn't get accounted for in the cost calculation or anywhere in the planner.

The only solution I can think of for this is that `LockMutexGroup` needs to forcibly override the start points used by the planner for as long as it's active. I'll try that out and see if it resolves this scenario.

xiyuoh (dismissed):
Thanks for nailing down the issue and adding the fix! Tested this a bunch of times and it consistently works now, all 5 robots are able to move to their respective chargers successfully. It'd be very meaningful to get this fix in.

PR #491, opened 2025-12-09
Attempt to fix the spam generated by false completion of a task during detours in GoToPlace

arjo129:
### Fixed bug

https://github.com/open-rmf/rmf_demos/issues/340

### Fix applied

Exactly what @mxgrey suggested.

### GenAI Use

- [x] I did not use GenAI

Generated-by:

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_RobotUpdateHandle.hpp
@@ -167,32 +167,13 @@ public:
       if (!context)
         return;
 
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
```diff
@@ -745,12 +746,15 @@ void GoToPlace::Active::_execute_plan(
   }
   else
   {
+    auto event = rmf_task::events::SimpleEventState::make(_assign_id->assign(), "detour", "", Status::Underway, {});
```
mxgrey: Instead of a blank string for the description, maybe we can put something like `"the robot is parking until its destination becomes available"`.

PR #498, opened 2025-12-31
Accurate GoToPlace task update and reservation fixes

xiyuoh:
This PR addresses https://github.com/open-rmf/rmf_demos/issues/342 with the following fixes:
- Ensures that the task status is reflected correctly during reservation-related go to place tasks
- Check whether the requester is already sitting on a valid parking spot before reporting that there are no free waitpoints available

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/ReservationManager.cpp
@@ -33,6 +33,11 @@ void ReservationManager::replace_ticket(
   {
     if (new_allocation.ticket.ticket_id != _allocation->ticket.ticket_id)
     {
+      // Do not publish release request if this is a re-allocation
+      if (new_allocation.resource == _allocation->resource)
+      {
+        return;
+      }
       RCLCPP_INFO(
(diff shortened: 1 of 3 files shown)


arjo129:
Can we run [these tests](https://github.com/arjo129/stress_test_rmf_chope) once to make sure the other behaviors are correct (You have to manually watch and run them). If possible we should add a test to that repo that replicates this bug.

arjo129 (approved):
Thanks for the changes and debugging this very hairy stuff.

I ran this with the office world and our test cases:
- :heavy_check_mark:  `test_multi_place.bash`
- :heavy_check_mark:  `test_swap.bash`
- :heavy_check_mark:  `test_cancellation.bash`
- :heavy_check_mark:  `test_basic.bash`
- :heavy_check_mark:  `test_current_waitspot.bash`
- :heavy_check_mark: `test_emergency_pullover`
- :cloud_with_rain:   `test_patrol.bash` - Very flaky, mostly the traffic planner can deadlock. Hard to reliably reproduce.

xiyuoh:
Are the flaky traffic scenarios caused by reservation mixups or negotiation failures? I notice the latter sometimes, and it seems to be tied to sim rate

Issue #493, opened 2025-12-09
[Other issue]: Path replan is interrupted during emergency


lttey:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf/open-rmf-general/103).

### Description

By running office simulation, i realized path replan is not triggered if lane closure triggered during emergency. Further looking into [GoToPlace::_find_plan()](https://github.com/open-rmf/rmf_ros2/blob/bcf0e3aa992763cc25670f8158ae355ac6b6aea9/rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp#L531), it's interrupted from proceeding to plan path.

It can be referred to the screencast below where robot actually stopped before the closed lane (from wp, close to patrol_D2) , even though there's alternate route which is working in non-emergency scenario.  

https://github.com/user-attachments/assets/387057fa-3cb0-4250-9932-7dd25b9818b4

We need to change path of robot which is retreating to parking lot during emergency, then encountered such issue on replan. Wonder it's intended behaviour which prevents robot from replanning its path during emergency? What's the suggestion to enable replan in such scenario?

mxgrey:
> To solve this, wonder we should have similar [replan subscriber](https://github.com/open-rmf/rmf_ros2/blob/bcf0e3aa992763cc25670f8158ae355ac6b6aea9/rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp#L186) in EmergencyPullover, so as to avoid the interrupt.

Yes, I believe you've identified the correct solution! It seems we coincidentally haven't encountered replans during emergency pullovers until now, so the emergency pullover implementation was simply missing that subscriber.

PR #500, opened 2026-01-07
enable path replanning during emergency pull over

lttey:

As refer to issue [493](https://github.com/open-rmf/rmf_ros2/issues/493)
Path re-planning is enabled for EmergencyPullover by creating second subscriber to replan_request under EmergencyPullover::Active::make

### GenAI Use

- [/] I did not use GenAI

Generated-by:


mxgrey (approved):
Sorry for the delay in reviewing.

This looks like exactly the right fix, thanks for the submission!

PR #511, opened 2026-03-10
Clean up LegacyPhaseShim subscription after task_summaries removal

### LINKED ISSUE #477: [Bug]: Remove publishers related to /task_summaries topic

### OS and version

Ubuntu 24.04

Source build

_No response_

main

### ROS distribution

Jazzy

### ROS installation type

Source build

_No response_

_No response_

### Description of the bug

The /task_summaries topic is not publishing any data despite having an active publisher. This critical issue prevents task monitoring and visualization systems from receiving task status updates from fleet adapters.

1. Launch the 4D demo: ros2 launch rmf_demos_gz office.launch.xml
2. Wait for all components to start (fleet adapter, traffic schedule, etc.)
3. Check if the topic exists: ros2 topic list | grep task_summaries
4. Verify publisher info: ros2 topic info /task_summaries
5. Attempt to read messages: ros2 topic echo /task_summaries --once
6. Check publishing rate: ros2 topic hz /task_summaries
7. Submit a delivery task to trigger task execution
8. Observe that no messages are published to the topic

### Expected behavior

- The /task_summaries topic should publish rmf_task_msgs/msg/TaskSummary messages containing:

1. Task status updates (QUEUED, ACTIVE, COMPLETED, FAILED, etc.)
2. Task progress information
3. Robot assignment details
4. Task timing information

- Messages should be published when:

1. 
2. Tasks are submitted to robots
3. Task states change
4. Tasks complete or fail
5. Robots are assigned to tasks

-  The topic should show a publishing rate > 0 Hz when tasks are active

### Actual behavior

1. The/task_summaries topic exists with 1 publisher and 0 subscribers
2. No messages are published to the topic
3. ros2 topic echo /task_summaries --once produces no output
4. ros2 topic hz /task_summaries shows no publishing rate
5. Task monitoring and visualization systems receive no data
6. The topic remains silent even when tasks are submitted and executed

RMF Version: Built from source (commit: fa98228684dad6d75971bdd5818af320fadbcc0d)
ROS2 Distribution: Jazzy

###Investigation results
$ ros2 topic info /task_summaries
Type: rmf_task_msgs/msg/TaskSummary
Publisher count: 1
Subscription count: 0

$ ros2 topic echo /task_summaries --once
No output

$ ros2 topic hz /task_summaries
No output

##Code Analysis:
Publisher is created in src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp:61-63
Topic name defined in src/rmf/rmf_ros2/rmf_fleet_adapter/include/rmf_fleet_adapter/StandardNames.hpp:54
Issue appears to be in the connection between LegacyTask status updates and the ROS publisher
##Impact:
Task monitoring systems cannot track robot progress
Visualization dashboards receive no task data
External integrations fail
Debugging and troubleshooting become difficult
##Related Files:
src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/LegacyTask.cpp
src/rmf/rmf_ros2/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp

Matthew0803:
## Bug fix
#477 
### Fixed bug

The robot fleet system used to publish a "task status bulletin board" (/task_summaries topic) so other parts of the system could watch what robots were doing. That bulletin board was removed because nobody was reading it. But leftover code in LegacyPhaseShim still used the old bulletin board message format to figure out when a robot task started, finished, or failed, instead of just knowing those things directly. 

### Fix applied

Simplify the subscription code to:

1. Immediately mark the event as "Underway" when the phase begins (like all other events in the codebase do)
2. Only use incoming messages for logging and failure detection
3. Let the completion handler properly set the "Completed" status
4. Remove the now-unnecessary _last_state_value tracking field


mxgrey (changes requested):
I can understand having a default switch to `Underway` status at the start and a switch to `Completed` status at the end if the task accidentally finished with an `Underway` status, but I don't understand why would discard other status changes that happen in between. I've put more detailed comments inline.

mxgrey (approved):
I've filtered this PR down to the key contributions and reverted what I felt was an unnecessary change in the way task statuses were being converted. I think this is good to merge now.

luca-della-vedova (commented):
I am not really familiar with this part of the codebase but the reason behind my TODO is that there is a large chunk below that looks like a subscription callback for `TaskSummary` messages, it sounds like once we removed all publishers on that topic this effectively became dead code?

mxgrey:
@luca-della-vedova 

> the reason behind my TODO is that there is a large chunk below that looks like a subscription callback for TaskSummary messages, it sounds like once we removed all publishers on that topic this effectively became dead code?

We're actually using the `TaskSummary` message still as an interface between these "legacy" task phase implementations and the `LegacyPhaseShim`. The messages don't get published on ROS, but they get streamed out by the legacy task phases via rxcpp, and then observed by the `LegacyPhaseShim`, which then converts the `TaskSummary` data into the "Task V2" state information.

PR #514, opened 2026-04-07
Resolve compilation warnings

aaronchongth:
### Fixed bug

Taking inspiration from https://github.com/open-rmf/rmf_traffic/pull/132, I performed source builds of Open-RMF and asked the agent to list out compilation warnings with easy and non-intrusive fixes throughout the code base.

More PRs will be opened in other repositories.

### Fix applied

* typos, syntax issues, re-ordering of initialization
* missing references, unused parameters or variables
* unnecessary `std::move` calls
* uninitialized members
* instead of using variables that have been marked `deprecated`, use string literals instead
* replacing the use of deprecated `.delay(...)` with `.cumulative_delay(...)`, as mentioned in https://github.com/open-rmf/rmf_traffic/blob/d5de7f3a982bd503a7424255a74362337b1fe584/rmf_traffic/include/rmf_traffic/schedule/Participant.hpp#L80
  * this change was slightly more involved, so should be reviewed carefully, just in case I am misusing the `.cumulative_delay(...)` API

I've skipped any of the required changes that were too intrusive to reduce the compiler warnings.

Overall stats regarding compilation warnings that were not related to thirdparty libraries or vendored libraries,
* compilation warnings on main: ~187
* after fixes throughout codebase: ~86
* reduced 101 warnings (-54%)

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Claude Sonnet 4.6

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
@@ -76,7 +76,7 @@ std::shared_ptr<Node> Node::make(
 
   node->_target_emergency_notice_obs =
     node->create_observable<TargetEmergencyNotice>(
-    "emergency_signal", transient_qos);
+    rmf_traffic_ros2::EmergencySignalTopicName, transient_qos);
 
   node->_ingestor_request_pub =
     node->create_publisher<IngestorRequest>(
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
@@ -72,11 +72,11 @@ std::shared_ptr<Node> Node::make(
 
   node->_emergency_notice_obs =
     node->create_observable<EmergencyNotice>(
-    rmf_traffic_ros2::EmergencyTopicName, transient_qos);
+    "fire_alarm_trigger", transient_qos);
 
   node->_target_emergency_notice_obs =
     node->create_observable<TargetEmergencyNotice>(
-    rmf_traffic_ros2::EmergencySignalTopicName, transient_qos);
+    "emergency_signal", transient_qos);
 
   node->_ingestor_request_pub =
     node->create_publisher<IngestorRequest>(
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
@@ -72,7 +72,7 @@ std::shared_ptr<Node> Node::make(
 
   node->_emergency_notice_obs =
     node->create_observable<EmergencyNotice>(
-    "fire_alarm_trigger", transient_qos);
+    rmf_traffic_ros2::EmergencyTopicName, transient_qos);
 
   node->_target_emergency_notice_obs =
     node->create_observable<TargetEmergencyNotice>(
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/DoorOpen.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/DoorOpen.cpp
@@ -128,16 +128,23 @@ void DoorOpen::ActivePhase::_init_obs()
           // supervisor sees our request.
           me->_publish_open_door();
 
+          const auto plan_id =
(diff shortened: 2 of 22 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
```diff
@@ -72,11 +72,11 @@ std::shared_ptr<Node> Node::make(
 
   node->_emergency_notice_obs =
     node->create_observable<EmergencyNotice>(
-    rmf_traffic_ros2::EmergencyTopicName, transient_qos);
+    "fire_alarm_trigger", transient_qos);
 
   node->_target_emergency_notice_obs =
     node->create_observable<TargetEmergencyNotice>(
-    rmf_traffic_ros2::EmergencySignalTopicName, transient_qos);
+    "emergency_signal", transient_qos);
```
xiyuoh: I'd recommend reverting this, so that [StandardNames.hpp](https://github.com/open-rmf/rmf_ros2/blob/main/rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp) is the source of truth for topic names

rmf_fleet_adapter/src/rmf_fleet_adapter/phases/DoorOpen.cpp
```diff
@@ -128,16 +128,23 @@ void DoorOpen::ActivePhase::_init_obs()
...
           if (delay > std::chrono::seconds(0))
           {
             me->_context->worker().schedule(
-              [context = me->_context, delay](const auto&)
+              [context = me->_context, plan_id, current_cumulative, delay](
+                const auto&)
               {
-                context->itinerary().delay(delay);
+                context->itinerary().cumulative_delay(
+                  plan_id, current_cumulative + delay);
```
xiyuoh: We probably don't need the extra calculations in between if we're just updating the cumulative delay.

```suggestion
          const auto new_cumulative = me->_context->now() - me->_expected_finish;
          if (new_cumulative > std::chrono::seconds(0))
          {
            me->_context->worker().schedule(
              [context = me->_context, plan_id, new_cumulative](
                const auto&)
              {
                context->itinerary().cumulative_delay(
                  plan_id, new_cumulative);

xiyuoh (commented):
Thanks for the cleanup! The `cumulative_delay()` changes make sense to me. Left some minor comments, otherwise LGTM

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Node.cpp
aaronchongth: revert in https://github.com/open-rmf/rmf_ros2/pull/514/commits/34deb8afff294a290a8ee415ceb3f67271267d26

split up lines in https://github.com/open-rmf/rmf_ros2/pull/514/commits/0ccd47b6541e3890b943006917b8c6a5419b0159, to not include `EmergencySignalTopicName` in the deprecated tag

rmf_fleet_adapter/src/rmf_fleet_adapter/phases/DoorOpen.cpp
aaronchongth: https://github.com/open-rmf/rmf_ros2/pull/514/commits/e2efb9dadafdc7f40894d2c81334994064398fa5, I just did the addition outside and captured a single variable

I believe the math is still required,

`delay` in old code is not equivalent to `new_cumulative` in your suggestion, so behaviors will change

xiyuoh: `delay` is only used to check if the cumulative delay changed though, which seems like an unnecessary calculation. The difference is very minor, I'm not super particular, just that `delay` and `current_expected_finish` is not required at all here. I might suggest just doing `if (new_cumulative != current_cumulative)` instead of calculating the above. Will leave it to you.

xiyuoh (approved):

rmf_fleet_adapter/src/rmf_fleet_adapter/phases/DoorOpen.cpp
aaronchongth: `expected_finish` does not get updated throughout the phase, only `cumulative_delay` gets updated as time goes by. So comparing just the current time and expected finish does not take into account of increments of delay.

We can chat more if needed. I'll just merge these changes, since the whole goal is not to change any behaviors

PR #517, opened 2026-04-23
Force C++-17 for rmf_websocket

luca-della-vedova:
### Fixed bug

Fix builds of `rmf_websocket` for rolling by forcing it to build under C++-17.

### Fix applied

Discussion on Zulip [#Open-RMF PMC (Restricted posting) > rmf_websocket broken on Rolling](https://openrobotics.zulipchat.com/#narrow/channel/526047-Open-RMF-PMC-.28Restricted-posting.29/topic/rmf_websocket.20broken.20on.20Rolling/with/590356684)
ROS with Lyrical is moving to default to C++-20. This doesn't work for `rmf_websocket` since the underlying library does not compile with C++-20 (there is an [open PR](https://github.com/zaphoyd/websocketpp/pull/1182) for this to be fixed but the upstream repository doesn't see much activity.
I just changed the build to enforce C++-17.

### Test it!

Just try to build the library with C++-20 (which is what ament_cmake_ros package will do by default from now on, _I believe_):

```
$ colcon build --packages-up-to rmf_websocket --cmake-args -DCMAKE_CXX_STANDARD=20 -DCMAKE_CXX_STANDARD_REQUIRED=ON
```

It should fail before this PR, succeed afterwards.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/rmf_websocket/CMakeLists.txt
+++ b/rmf_websocket/CMakeLists.txt
@@ -2,9 +2,8 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rmf_websocket)
 
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 17)
-endif()
+# Websocketpp is not compatible with CXX above 17
+set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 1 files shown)


mxgrey (approved):
Thanks for catching this!

Issue #515, opened 2026-04-12
[Other issue]: Should expected_finish_state() reflect cumulative delay from blocked/still-waiting phases?


kjchee:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf/open-rmf-general/103).

### Description

When a robot is blocked/delayed in a phase (e.g. [LockMutexGroup](https://github.com/open-rmf/rmf_ros2/blob/9d52b946a24bb7b689a1d532b0c31639f5fc97fa/rmf_fleet_adapter/src/rmf_fleet_adapter/events/LockMutexGroup.cpp#L347-L351) apply delay while waiting for a mutex, [RequestLift](https://github.com/open-rmf/rmf_ros2/blob/9d52b946a24bb7b689a1d532b0c31639f5fc97fa/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/RequestLift.cpp#L204-L216) apply delay while waiting for a stuck lift), [TaskManager::expected_finish_state()](https://github.com/open-rmf/rmf_ros2/blob/9d52b946a24bb7b689a1d532b0c31639f5fc97fa/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp#L853-L876) continues to return the original optimistic finish time computed at task assignment. It does not reflect the delay that has accumulated while the phase waits.

This can cause the task planner to prefer a blocked robot over another robot when assigning new tasks, because the blocked robot's reported finish time may be in the past (but in fact the robot is still stuck/waiting in the phase).

I wanted to check whether this is an intentional design choice or an oversight before opening a PR. Is `TaskManager::expected_finish_state()` intended to be a fixed "planned outcome" anchor, or should it reflect the robot's actual progress?

eg. to reflect robot's actual progress:
```cpp
auto TaskManager::expected_finish_state() const -> State
{
  std::lock_guard<std::recursive_mutex> lock(_mutex);
  if (!_direct_queue.empty())
  {
    rmf_task::State return_state = _direct_queue.rbegin()->assignment.finish_state();
    return_state.idle(false);
    return return_state;
  }

  if (_active_task)
  {
    rmf_task::State return_state = _context->current_task_end_state();
    ////////////--change here--////////////////////////////
    // Adjust for cumulative delay so that phases blocked on external
    // resources report a realistic finish time to the task planner.
[...]

PR #518, opened 2026-04-24
Make expected_finish_state reflect cumulative delay

kjchee:
### Improved feature

This improvement addresses: #515.

`TaskManager::expected_finish_state()` is now delay-aware, and `GoToPlace::remaining_time_estimate()` includes a parabolic frustration penalty during long stalls.

### Improvement description
In existing RMF, a robot stuck at a mutex/lift/door can still be picked for new tasks even though it has a long delay in its  active task.

This happens because `TaskManager::expected_finish_state()` returns the optimistic dispatch-time finish, which drifts into the past during stalls, so the task planner treats the stuck robot as "nearly free" and available. See the issue thread for the full discussion and design decisions.

Summary of changes:
- Route `TaskManager::expected_finish_state()` through a new `ActiveTask::estimate_remaining_time()` accessor that delegates to `Task::Active::estimate_remaining_time()`. The `_active_task` branch returns `(now + remaining)`, while the `_direct_queue` branch shifts queued finish times forward by the active task's accumulated delay.
- Add a parabolic frustration penalty (`delay² / reference_delay`, `reference_delay = 60s`) in `GoToPlace::remaining_time_estimate()` so the estimate keeps growing during long stalls.

#### Setup

1. Use this modified office map for testing: https://github.com/kjchee/rmf_demos/blob/fix/expected-finish-state-delay-aware/rmf_demos_maps/maps/office/office.building.yaml

    fix_task_end_state

2. Set `finishing_request: nothing` so the robot stays where it is during launch (otherwise tinyRobot1 will be sent to `supplies`).
3. Manually turn off `mutex_group_supervisor` by commenting it out in `common.launch.xml` to let the robot get stuck at the mutex.

#### Steps (same for Before/After)

1. Launch office:
   ```bash
   ros2 launch rmf_demos_gz office.launch.xml
   ```
2. Send a patrol task to waypoint `patrol_D1`:
   ```bash
   ros2 run rmf_demos_tasks dispatch_patrol -p patrol_D1 --use_sim_time
   ```
3. RMF assigns the task to `tinyRobot1`.
4. When `tinyRobot1` reaches waypoint `mutex_entry` (and gets stuck), send a 2nd patrol task to waypoint `pantry`:
   ```bash
   ros2 run rmf_demos_tasks dispatch_patrol -p pantry --use_sim_time
   ```
5. RMF assigns the 2nd task to `tinyRobot1` as well (because `expected_finish_state()` is still optimistic at this point).
6. Wait for the next `reassign_task_interval` (default 120s) and observe the reassignment behavior.

#### Before fix

At step 6: the 2nd task remains assigned to `tinyRobot1` even though it is still stuck at the mutex entry.

#### After fix

At step 6: the 2nd task is reassigned to `tinyRobot2` (even though `tinyRobot2` is physically further from `pantry`) because `tinyRobot1`'s accumulated delay + frustration penalty now makes it look more expensive to the task planner.

### GenAI Use

- [x] I did not use GenAI

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
@@ -349,7 +349,22 @@ rmf_traffic::Duration GoToPlace::Active::remaining_time_estimate() const
     if (_execution->plan_id)
     {
       if (const auto delay = itin.cumulative_delay(*_execution->plan_id))
-        return finish - now + *delay;
+      {
+        // Parabolic frustration penalty (delay^2 / reference_delay):
+        // zero at zero delay, matches delay at reference_delay,
+        // grows much faster beyond.
+        // Used so remaining_time_estimate keeps growing with the stall
+        // instead of staying flat, which makes the task planner less likely
+        // to assign new tasks to robots stuck on lift/door/mutex events.
+        const auto reference_delay = std::chrono::seconds(60);
+        const double delay_s =
+          std::max(0.0, rmf_traffic::time::to_seconds(*delay));
+        const double reference_delay_s =
+          rmf_traffic::time::to_seconds(reference_delay);
+        const auto frustration_penalty =
+          rmf_traffic::time::from_seconds(delay_s * delay_s / reference_delay_s);
+        return finish - now + *delay + frustration_penalty;
+      }
     }
     else
     {
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -288,6 +288,14 @@ const std::string& TaskManager::ActiveTask::id() const
   return _task->tag()->booking()->id();
 }
 
+//==============================================================================
+rmf_traffic::Duration TaskManager::ActiveTask::estimate_remaining_time() const
+{
+  if (!_task)
+    return rmf_traffic::Duration(0);
+  return _task->estimate_remaining_time();
+}
+
 namespace {
 
 //==============================================================================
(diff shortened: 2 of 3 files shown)


xiyuoh (approved):
Ran the test case from the description and it worked great! I'm able to observe the parabolic cost taking effect.

If I understood the conversation from the issue ticket correctly, the implemented solution seems aligned, and leaves the switch to [_execution->sequence->remaining_time_estimate()](https://github.com/open-rmf/rmf_task/blob/66b2947f76d87db4e139c9639cabfca6e0dc356d/rmf_task_sequence/include/rmf_task_sequence/Event.hpp#L102) in follow-ups, pending implementing the logic in relevant events and phases.

Leaving this PR unmerged in case anyone else would like to take another pass.

rmf_fleet_adapter/src/rmf_fleet_adapter/events/GoToPlace.cpp
```diff
@@ -349,7 +349,22 @@ rmf_traffic::Duration GoToPlace::Active::remaining_time_estimate() const
...
       if (const auto delay = itin.cumulative_delay(*_execution->plan_id))
-        return finish - now + *delay;
+      {
+        // Parabolic frustration penalty (delay^2 / reference_delay):
+        // zero at zero delay, matches delay at reference_delay,
+        // grows much faster beyond.
+        // Used so remaining_time_estimate keeps growing with the stall
+        // instead of staying flat, which makes the task planner less likely
+        // to assign new tasks to robots stuck on lift/door/mutex events.
+        const auto reference_delay = std::chrono::seconds(60);
```
mxgrey, line 359: We should probably make this parameterized so users can set it to a delay threshold that makes sense for their system, but we can start with this as-is and do a follow-up PR when needed.

mxgrey (approved):
This looks great, thanks @kjchee ! I left one remark, but we can move forward with this as-is.

PR #520, opened 2026-05-06
Lyrical compatibility

mxgrey:
This PR updates `rmf_ros2` to be compatible with the current state of the ROS Lyrical distro.

This PR depends on https://github.com/open-rmf/nlohmann_json_schema_validator_vendor/pull/22

A major change introduced by this PR is to copy a fork of `websocketpp` directly into this repo. While this is a terrible practice, we are in a bind because the upstream source of `websocketpp` is not being maintained. The version of `websocketpp` that is distributed in Ubuntu 26.04 cannot be compiled with the version of Boost that is distributed in Ubuntu 26.04. There are [pull requests that have been opened upstream](https://github.com/zaphoyd/websocketpp/pull/1190) to fix these problems, but those PRs have gone unreviewed and unmerged.

### DIFF
--- a/.github/workflows/asan.yaml
+++ b/.github/workflows/asan.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   asan_test:
     name: rmf_ros2 asan
-    uses: open-rmf/rmf_ci_templates/.github/workflows/reusable_build.yaml@main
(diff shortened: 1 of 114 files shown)


luca-della-vedova (commented):
If we are vendoring websocketpp we should also remove it from the `rmf_websocket` package.xml to make sure we don't install it anymore

PR #521, opened 2026-05-14
Fix descrption typo in robot_state_aggregator launch file

Lots-ninety-nine:
The `buddy_subns` arg declaration in this launch file uses `descrption=` instead of `description=`. ROS 2 `launch_xml` rejects unknown attributes, so the file fails to load:

```
ValueError: Unexpected attribute(s) found in `arg`: {'descrption'}
```

Verified on `ros:humble` and `ros:jazzy`: after renaming, `ros2 launch <file> -s` prints arguments normally.


aaronchongth (approved):
thanks for catching the typo

aaronchongth:
merging as build issue on rolling has nothing to do with fix

PR #524, opened 2026-05-26
Fix cumulative delay calculation in EasyTrafficLight

kjchee:
### Fixed bug
During development of traffic light adapter, i observed there is some strange schedule marker lagging or flickering behavior.

Eg. videos below (`tinyRobot2` is a `trafficlight` robot, while `tinyRobot1` is a `fullcontrol` robot):

1. look at timestamp starting from 50s:

https://github.com/user-attachments/assets/2d48ae5f-8301-4c8f-9ab5-8d3568e66e48

2. look at timestamp starting from 24s:

https://github.com/user-attachments/assets/1b65a96f-6a9f-4d6e-94dc-5e0958597bb6

Then I tried to traced the `cumulative_delay` update and found that the there is an issue at the cummulative_delay calculation at `EasyTrafficLight::Implementation::Shared::update_delay:`
https://github.com/open-rmf/rmf_ros2/blob/9dedd0f746e5006d068237e78ca9dfbe47edf622/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyTrafficLight.cpp#L533

Here, `new_cumulative_delay` is supposed to be the "absolute total accumulated delay" starting from original plan's timestamp. But `hooks.node->rmf_now() - expected_time` is actually not `cumulative_delay` but a "delta delay". It is the current time minus the trajectory's already-shifted time where `expected_time` is the already-shifted time.

### Fix applied
So we will need to correct `new_cumulative_delay` to `existing_cumulative_delay + delay_delta`.


mxgrey (commented):
Thanks for the fixes!

mxgrey (approved):

PR #525, opened 2026-05-26
Fix EasyTrafficLight publish fleet state

jixiancgh:
### Fixed bug

While testing with our own traffic light robot using the EasyTrafficLight module, we found some issues: 
- EasyTrafficLight never called `fleet_state_pub->publish()` in `publish_fleet_state()`, so no FleetState messages were emitted. 
- The location field used the static graph waypoint coordinates, which always reflect the next waypoint rather than the robot's actual mid-segment position.

### Fix applied

- Added the missing `publish()` call in `publish_fleet_state()`.
- Changed the location source to `last_known_location` (with fallback to the graph waypoint if no update has arrived yet), so the published position reflects the robot's real-time pose.

### Known limitation

Each robot currently publishes its own FleetState containing only itself. With multiple robots in a fleet, subscribers receive multiple single-robot messages instead of one aggregated fleet message. A proper fix requires moving fleet level publishing to a shared context, we can open this in a separate issue.

### GenAI Use

- [x] I did not use GenAI

Generated-by:


fcarmone:
Hi, I encountered the same issue recently. After applying your fix (and #524), I'm still unable to get the robot to send real-time position data.

I suspect there may be a logical error in my implementation. Could you describe what the expected position data workflow looks like? Alternatively, if you started from an existing template, could you point me to it?

Thanks in advance.

kjchee:
Hi @fcarmone ,
> if you started from an existing template, could you point me to it?

We are now developing a new python template using the existing python binding.


> I suspect there may be a logical error in my implementation. Could you describe what the expected position data workflow looks like?


Hmmm, for this it will be hard to identify whats wrong because we don't know your implementation (eg. real robot or slotcar, or using python API or cpp and how you trigger each API etc...).


But for the isseu on "unable to get the robot to send real-time position data", it boils down to how you update the traffic light core with your robot's real-time location while triggering the API, eg:

- `moving_from(checkpoint, pose)`
- `waiting_after(checkpoint, pose)`

You can check on the `pose` you passed in, is it the real-time robot pose (not the checkpoint coordinate)? Also, have you triggered `moving_from(checkpoint, pose)` while your robot is moving (1hz for our case), and `waiting_after` while it's paused and waiting in between checkpoints? 

You can debug from this direction. Hope this helps!

mxgrey (approved):
Thanks for these fixes!

mxgrey:
@jixiancgh You'll need to git merge the latest `main` branch of https://github.com/open-rmf/rmf_ros2 into your branch for this PR or else the CI for lyrical and resolute will not be able pass. Once that's merged and the CI passes, we'll merge this PR.

jixiancgh:
@mxgrey I've merged the latest `main` branch.

PR #541, opened 2026-08-08
Support both JSON schema validator target names

wolfv:
Different releases of `nlohmann_json_schema_validator` export different CMake target names. The version pinned by `nlohmann_json_schema_validator_vendor` exports the unnamespaced `nlohmann_json_schema_validator` target, while newer releases provide `nlohmann_json_schema_validator::validator`.

Select the target that is actually available in both `rmf_websocket` and `rmf_task_ros2`. This fixes clean builds against the currently vendored validator without dropping compatibility with newer validator packages.

The unnamespaced target is checked first because `rmf_task_ros2` may see an unresolved namespaced target reference transitively from `rmf_websocket`.

Signed-off-by: Wolf Vollprecht <w.vollprecht@gmail.com>

### DIFF
--- a/rmf_task_ros2/CMakeLists.txt
+++ b/rmf_task_ros2/CMakeLists.txt
@@ -27,6 +27,12 @@ find_package(nlohmann_json REQUIRED)
 find_package(nlohmann_json_schema_validator_vendor REQUIRED)
 find_package(nlohmann_json_schema_validator REQUIRED)
 
+if(TARGET nlohmann_json_schema_validator)
+  set(json_schema_validator_target nlohmann_json_schema_validator)
+else()
+  set(json_schema_validator_target nlohmann_json_schema_validator::validator)
+endif()
+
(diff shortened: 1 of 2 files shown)


aaronchongth (commented):
Interesting, is this fix required if built against `main` of https://github.com/open-rmf/nlohmann_json_schema_validator_vendor, where the version is 2.4, but not the lyrical branch?

I just managed to build off of `main` for `rmf_websocket` and `rmf_task_ros2`, making sure that `nlohmann_json_schema_validator_vendor` was `main` with 2.4. Can you provide some more information or ways to reproduce this issue?

mxgrey (approved):
This PR is a bit questionable because we do not generally support mixing versions of Open-RMF packages that target different ROS distros. That being said, I don't see any specific harm in adding this alias for the target. While the configuration error could be a clue to the user that they have a package version mismatch, most users will not recognize the cause of the configuration error, so it will seem like there's simply a bug in the build system of the `rmf_ros2` repo.

Overall, if this reduces a source of confusion for end users, then I see no point in blocking it.

Issue #544, opened 2026-08-19
ScheduleNode::request_changes never fills response->node_id — empty identity makes MirrorManager silently reset its mirror on every remedial request, so a mirror that misses one patch can churn forever


casta42:
## Summary

`ScheduleNode::request_changes` (`rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Node.cpp`,
Humble branch, handler starting ~L1029) populates `response->result` /
`response->error` but never sets `response->node_id`. The response therefore
carries a default-constructed `ScheduleIdentity` (`node_uuid: ''`,
`timestamp: 0`). Compare `register_query`, which does set it
(`response->node_id = node_id;`, ~L720), and the mirror-update messages
(~L896).

On the mirror side, `MirrorManager::Implementation::validate_service_response`
sees a node UUID that differs from the one it stored. `need_reconnection()`
requires the incoming identity to be NEWER before reconnecting, and a
zero timestamp can never be newer — so the code takes the "this response
came from an old schedule node" branch: it calls `reconnect_services()`,
which **silently calls `mirror->reset()`** (latest version -> nullopt) and
rejects the response.

Net effect: the recovery path poisons the mirror it is trying to heal.

## Failure loop (observed, ROS 2 Humble, CycloneDDS 0.10.5, single host)

1. A mirror misses one incremental patch (any transient cause — busy boot,
   load burst). It calls `request_changes`.
2. The schedule node queues the remedial (topic-side, correct identity) AND
   returns the service response with the EMPTY identity.
3. Two things race into the mirror's executor: the remedial update (topic)
   which would fully restore it, and the service response which resets it.
4. If the response is processed last, the mirror is left at version `none`;
   the next incremental patch fails
   (`Failed to update using patch ... (mirror version: none, ...)`);
   the update timeout fires; goto 1. We observed this loop continuing
   indefinitely (hours) at exactly the update-timeout cadence, with BOTH
   in-process mirrors (fleet adapter + schedule visualizer) locked out
   while the traffic planner ran blind.
5. Whether a deployment ever sees this depends purely on executor/transport
   timing: with heavy CycloneDDS tracing enabled (which slows delivery), the
   same stack is stable for 20+ minutes; untraced it locks up within ~5
   minutes of our 6-robot sim starting. A host-side ROS core apt sync (with
   the RMF workspace unchanged) was enough to flip our deployment from
   never-seeing-it to always-seeing-it.

## Evidence

[...]

PR #545, opened 2026-08-19
Backport to humble

### LINKED ISSUE #537: ScheduleNode::request_changes never sets response->node_id — fleet-adapter mirrors reset to 'mirror version: none' whenever the schedule idles >20s
## Summary

`ScheduleNode::request_changes` never sets `response->node_id`, so every `RequestChanges` service response carries a default-initialized `ScheduleIdentity` (`node_uuid=""`, `timestamp=0`). On the client side, `MirrorManager::validate_service_response` treats that empty identity as a foreign schedule node and calls `reconnect_services()`, which resets the mirror.

The result: **any fleet adapter whose schedule goes quiet for longer than the mirror's 20 s update timer livelocks permanently** at:

```
[WARN] Failed to update using patch for DB version N (mirror version: none, patch base: M); requesting new update
[INFO] [rmf_traffic_ros2::MirrorManager::request_update] Requesting changes for query ID [1] since beginning of recorded history
```

The cycle: mirror bootstraps fine → schedule goes quiet → the 20 s update timer fires → `request_changes` is sent → **the ack itself wipes the mirror** → the next incremental patch fails against `mirror version: none` → a remedial re-bootstraps the mirror → quiet again → repeat forever. ~~`add_robot`'s handle callback context never stabilizes, and on an idle deployment the adapter never recovers.~~ **EDIT 2026-07-29:** this sentence conflated an unrelated adapter-side bug with this one. With that fixed, adapters come up and execute tasks normally even against an unpatched schedule — the cycle's real cost is the perpetual full-history remedial traffic and the empty-mirror windows, not a functional halt. A/B validation in the comments below.

## Where

`rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Node.cpp`, `ScheduleNode::request_changes` (~line 1031 on `humble`). Its sibling `register_query` (same file, ~line 720) *does* set `response->node_id`; `request_changes` was missed. The omission appears to date to ca462b5 (PR #232, "Robust schedule failover", 2022) and is still present on `main` as of 2026-07-28.

## Reproduction

ROS 2 Humble, stock debs (`ros-humble-rmf-traffic-ros2` 2.1.8-1jammy, arm64 or amd64), Docker Compose:

1. Run `rmf_traffic_schedule`, `rmf_task_dispatcher`, and one full-control fleet adapter (classic `Adapter` API, one simulated robot) with **no task churn** — an idle robot updating its state normally.
2. Wait ~20–30 s after the mirror bootstraps.
3. Observe the log signature above repeating every ~10–30 s, with the `since version [N]` value *climbing* — the tell that the mirror was alive right before each wipe (this is a bootstrap-then-wipe loop, not a bootstrap failure).

Direct wire proof, no instrumentation needed:

```bash
ros2 service call /rmf_traffic/request_changes rmf_traffic_msgs/srv/RequestChanges "{query_id: 1, version: 0, full_update: true}"
# response: node_id: {node_uuid: '', timestamp: {sec: 0, nanosec: 0}}   <-- empty identity
```

Two corroborations from our environment:

- Instrumented `MirrorManager` (log lines added at `handle_update` / `apply_patch` / after `mirror->update`) shows remedials received and applied successfully (`ok=1`, 150/151 patches over 120 s) — and the mirror version reset to none after each `request_changes` ack. The failure is not message delivery.
- Keeping the schedule DB changing at intervals **shorter than 20 s** permanently prevents the livelock (the update timer never fires, so `request_changes` is never sent). This matches deployments with constant task churn never noticing the bug.

## Fix (one line, A/B validated)

```cpp
void ScheduleNode::request_changes(
  const request_id_ptr& request_header,
  const RequestChanges::Request::SharedPtr& request,
  const RequestChanges::Response::SharedPtr& response)
{
  response->node_id = node_id;   // <-- missing; register_query already does this
  ...
```

With an overlay carrying only this change on the schedule node: zero mirror resets across many request cycles, including through a 32 s `docker pause` of the schedule container. Without it: a reset on every cycle, reproduced 100%.

## Why this can look like a platform issue

Two deployments dodge it accidentally, which may explain why it has been hard to pin down:

- `rmf_traffic_schedule_monitor` never drives timer-based `request_changes` against a quiet schedule the same way, so monitors keep working while fleet adapters livelock.
- Any deployment with steady task/robot churn keeps the DB version moving, the 20 s timer never fires, and the bug never triggers.

Issue #532 (closed as unreproducible/platform-specific) matches this fact pattern closely — same log signature, monitor unaffected — and may share this root cause.

## Environment

- `ros:humble-ros-base-jammy` (arm64, Docker Desktop on macOS; also reproduced with default `rmw_fastrtps_cpp` and `rmw_cyclonedds_cpp` — the bug is transport-independent)
- `ros-humble-rmf-traffic-ros2` 2.1.8-1jammy (June 2026 rebuild); code inspection confirms the omission on `main`

aaronchongth:
### Backport

* #524 
* #525 
* #538 

This resolves #537 #544 for humble, and also backports other fixes that do not have conflicts

### GenAI Use

- [x] I did not use GenAI

Generated-by:

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyTrafficLight.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyTrafficLight.cpp
@@ -522,6 +522,10 @@ void EasyTrafficLight::Implementation::Shared::update_delay(
   {
     const auto slices = state.current_itinerary_slice();
 
+    const auto existing_cumulative_delay =
+      state.itinerary->cumulative_delay(state.itinerary->current_plan_id())
+      .value_or(rmf_traffic::Duration(0));
+
     for (const auto& slice : slices)
     {
       try
@@ -530,7 +534,8 @@ void EasyTrafficLight::Implementation::Shared::update_delay(
           rmf_traffic::agv::interpolate_time_along_quadratic_straight_line(
           slice.trajectory(), location->block<2, 1>(0, 0));
 
-        new_cumulative_delay = hooks.node->rmf_now() - expected_time;
+        const auto delay_delta = hooks.node->rmf_now() - expected_time;
+        new_cumulative_delay = existing_cumulative_delay + delay_delta;
         break;
       }
       catch (const std::exception& e)
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyTrafficLight.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyTrafficLight.cpp
@@ -1085,7 +1085,8 @@ void EasyTrafficLight::Implementation::Shared::publish_fleet_state() const
     .location(std::move(location))
     .path({});
 
-  const auto& fleet_state = rmf_fleet_msgs::build<rmf_fleet_msgs::msg::FleetState>()
+  const auto& fleet_state =
+    rmf_fleet_msgs::build<rmf_fleet_msgs::msg::FleetState>()
     .name(fleet_name)
     .robots({std::move(robot_state)});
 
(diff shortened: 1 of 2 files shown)


aaronchongth:
Use rebase and merge, do not squash and merge

arjo129:
Sorry about that! I missed that. Do you want me to revert the previous merge?

arjo129 (approved):

PR #558, opened 2026-09-11
Support both JSON schema validator target names in rmf_fleet_adapter

h-wata:
## Summary

#541 added a fallback for installations that export the JSON schema validator as `nlohmann_json_schema_validator` rather than `nlohmann_json_schema_validator::validator`, but it covered only `rmf_task_ros2` and `rmf_websocket`. `rmf_fleet_adapter` links the same library and still hardcodes the namespaced name, so on those installations its CMake generate step fails:

```
CMake Error at CMakeLists.txt:110 (target_link_libraries):
  Target "rmf_fleet_adapter" links to:

    nlohmann_json_schema_validator::validator

  but the target was not found.
```

This applies the same pattern `rmf_websocket` already uses, covering both the library and the `BUILD_TESTING` target. Installations that export the namespaced target take the `else()` branch and are unaffected.

## Test plan

Built with colcon on ROS 2 Jazzy in a workspace whose `nlohmann_json_schema_validator_vendor` exports the unnamespaced target. Before this change `rmf_fleet_adapter` failed at the CMake generate step; after it, the package configures and builds cleanly. The other two packages already fixed by #541 are unchanged.

## GenAI use

- [ ] I did not use a GenAI tool in this PR.

We hit this while building `rmf_fleet_adapter` to pick up another fix. The investigation and this PR description were prepared with the assistance of an AI coding assistant (Claude).

### DIFF
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -51,6 +51,12 @@ foreach(pkg ${dep_pkgs})
   find_package(${pkg} REQUIRED)
 endforeach()
 
+if(TARGET nlohmann_json_schema_validator)
+  set(json_schema_validator_target nlohmann_json_schema_validator)
+else()
+  set(json_schema_validator_target nlohmann_json_schema_validator::validator)
+endif()
+
 # Backward compatibility for yaml-cpp 0.7 that exports a yaml-cpp target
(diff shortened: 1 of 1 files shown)


aaronchongth (approved):
Thanks for the fix. Overall, I echo the sentiment in https://github.com/open-rmf/rmf_ros2/pull/541#pullrequestreview-5031638421 that this is a little odd, but there is no reason to not support this.

